Issue #9, opened 2020-12-29
Add installation instructions


david-allison:
Consider a `doctor.sh` script

* Java
* NDK
* Python
  * and google protobuf compiler
* Rust
   * 4 Android targets    
   * no spaces in NDK path (Windows)
   * 2/3 targets for testing & warn on macOS
   * windows compiler for testing
   * Nightly for testing

TarekkMA:
Do you mean to make a script which:
- Check if Java, NDK, Python (protobuf lib) are installed

What does these steps mean:
- 4 Android targets
- 2/3 targets for testing & warn on macOS
- Nightly for testing
- And should we check a c compiler on widows?

david-allison-1:
Sorry! Was writing the instructions for myself. 

Please also feel free to mark where the below needs to be moved in the documentation

**4 Android targets**

- we build the `.so` for 4 different Android architectures:

https://github.com/david-allison-1/Anki-Android-Backend/blob/6c16f621080382923f287dff089e86a235e8026b/rsdroid/build.gradle#L39

These need to be installed as targets for Rust:

https://github.com/david-allison-1/Anki-Android-Backend/blob/6c16f621080382923f287dff089e86a235e8026b/.github/scripts/install_rust_targets.sh#L1-L5

**2/3 targets for testing**

In `Anki-Android`, we run a lot of unit tests under Robolectric: https://github.com/robolectric/robolectric

This runs a simulated Android runtime under the JVM.

The problem with this is that the Rust libraries are built for Android, not for your native PC OS. 

To fix this, we have `rsdroid-testing`, which is a `jar`, containing the rust code compiled as native libraries for Windows (`dll`), macOS (`dylib`), and Linux (`.so`)

So we need rust targets for these 

https://github.com/david-allison-1/Anki-Android-Backend/blob/6c16f621080382923f287dff089e86a235e8026b/.github/scripts/install_rust_targets.sh#L6-L9

**macOS**

To my knowledge, a rust target cannot legally be cross-compiled for macOS unless on a macOS computer, as it requires the macOS SDK, which is protected by a license: https://www.apple.com/legal/sla/docs/xcode.pdf

We're fine with this on GitHub, as we can use macOS on our CI to generate a full build.

[...]

PR #38, opened 2021-03-20
bump gradle wrapper version

TarekkMA:
I was facing issues running gradle from command line using the current wrapper version.

#9

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.1.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.3-all.zip


TarekkMA:
I don't know should I keep working on this PR or make multiple?

I prefer making it one PR until I can figure out the build system for this project and document any issues I may encounter

david-allison-1:
Totally up to you - this is still beta software, so let me know what you'd prefer doing, no need for anything to get in the way.

TarekkMA:
Let's make it one PR for simplifying the process of setting the environment.

mikehardy:
gradle is on 6.8.3 now

TarekkMA:
Should we make it match that? I did only bump it until it worked. Not sure if anything would be affected.

mikehardy:
My personal rules for dependency updates:

If changing dependency versions, always look for the release notes on the dependency and bump it to current stable. 

The only exceptions to moving to current stable is if there is some deprecation we need to handle that we cannot yet (then log a deprecation / forward-port issue for future) or if there is an upstream bug blocking adoption of current stable, at which point you subscribe to that bug (or log if if needed!) and you update to the most recent version possible with a link in the code to the upstream issue stating the version affected and that it can be increased when the issue is resolved

mikehardy:
There is an implication to your last part "Not sure if anything would be affected" - that's the point, you are already in there making changes, *you* are responsible for seeing if anything is affected. As long as you are already doing that, might as well go all the way to current right?

TarekkMA:
Yes, you're right that does make sense.

mikehardy (approved):
I approve these 2 changed characters :laughing:

TarekkMA:
I will work on the other PR It does have the same commits as this one.

Issue #51, opened 2021-04-01
Execute Rust-based rslib-bridge unit tests in CI


david-allison:
About to add these, and they're not yet set up in CI

dae:
This was solved when this code was moved into the Anki repo.

PR #52, opened 2021-04-01
feat: fix OOM via memory limit on DB row results

### LINKED ISSUE #14: Implement RustSupportSQLiteDatabase:isLowMemory to enable LimitOffsetSQLiteCursor
Requires `ActivityManager` or similar to detect: 
* Low RAM Device
* Current device has low RAM

Which requires a `Context` object.

`RustSupportSQLiteDatabase` is created by `RustSupportSQLiteOpenHelper`. 

We can inject a `Context` there (or an interface abstracting it) into `BackendFactory`. 

This feels easy/somewhat reasonable architecturally (only a couple of levels of indirection), but I'll hold off in case there are better options. It feels rather "enterprise" with the layers of abstraction, and there's probably a more clean way.

david-allison-1:
We now use a 2MB page size, the same as CursorWindow.sCursorWindowSize

We occasionally got OOMs on methods which returned unbounded data
eg. getting the field data for notes: "Check Database" crashed

To fix this, instead of saying to the Rust "we want 1000 rows max",
we say "we want max 2MB of data"

The calculation for the data is inexact - string length, and 8 bytes
for doubles/longs.
Hasn't been tested thoroughly, but seems to only be < ~7%
off the protobuf size for a string-only column (underestimate).

We measure the rust in-memory usage rather than the size of the
serialized protobuf

Measuring the serialized size wasn't researched, but was assumed to be
hard, as we would need to stream into a protobuf collection, and be
able to efficiently query the new size for each row we add.

Main changes:
StreamingProtobufSQLiteCursor - no longer use fixed size pages
dbcommand: allow access via an offset to the result set rather than
via pages

Adds: setDbPageSize to allow the change of the size for debugging
Adds: Unit tests for the rust - not yet executed in CI (#51)
rename: getPage -> getNextSlice

bumps `anki` commit to add field sqlite.proto#DbResponse:start_index

Fixes #14 (no longer necessary)
Fixes https://github.com/ankidroid/Anki-Android/issues/8178

### DIFF
--- a/rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/DatabaseIntegrationTests.java
+++ b/rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/DatabaseIntegrationTests.java
@@ -23,7 +23,6 @@ import android.database.sqlite.SQLiteException;
 import androidx.sqlite.db.SupportSQLiteDatabase;
 
 import net.ankiweb.rsdroid.ankiutil.DatabaseUtil;
-import net.ankiweb.rsdroid.database.StreamingProtobufSQLiteCursor;
 import net.ankiweb.rsdroid.database.testutils.DatabaseComparison;
 
 import org.junit.Test;
@@ -38,6 +37,10 @@ import static org.junit.Assert.fail;
 @RunWith(Parameterized.class)
 public class DatabaseIntegrationTests extends DatabaseComparison {
 
+    private static final int INT_SIZE_BYTES = 8;
+    private static final int OPTIONAL_BYTES = 1;
+    /** Number of integers in 1 page of DB results when under test (111) */
+    public static int DB_PAGE_NUM_INT_ELEMENTS = TEST_PAGE_SIZE / (INT_SIZE_BYTES + OPTIONAL_BYTES);
 
     @Test
     public void testScalar() {
@@ -254,12 +257,12 @@ public class DatabaseIntegrationTests extends DatabaseComparison {
 
         db.execSQL("create table test (id int)");
 
-        for (int i = 0; i < StreamingProtobufSQLiteCursor.RUST_PAGE_SIZE; i++) {
+        for (int i = 0; i < DB_PAGE_NUM_INT_ELEMENTS; i++) {
             db.execSQL("insert into test VALUES (1)");
         }
 
         try (Cursor c = db.query("select * from test")) {
-            assertThat(c.getCount(), is(StreamingProtobufSQLiteCursor.RUST_PAGE_SIZE));
+            assertThat(c.getCount(), is(DB_PAGE_NUM_INT_ELEMENTS));
         }
     }
 
@@ -269,12 +272,12 @@ public class DatabaseIntegrationTests extends DatabaseComparison {
 
         db.execSQL("create table test (id int)");
 
-        for (int i = 0; i < StreamingProtobufSQLiteCursor.RUST_PAGE_SIZE + 1; i++) {
+        for (int i = 0; i < DB_PAGE_NUM_INT_ELEMENTS + 1; i++) {
             db.execSQL("insert into test VALUES (1)");
         }
 
         try (Cursor c = db.query("select * from test")) {
-            assertThat(c.getCount(), is(StreamingProtobufSQLiteCursor.RUST_PAGE_SIZE + 1));
+            assertThat(c.getCount(), is(DB_PAGE_NUM_INT_ELEMENTS + 1));
         }
     }
 
--- a/rsdroid/src/main/java/net/ankiweb/rsdroid/BackendMutex.java
+++ b/rsdroid/src/main/java/net/ankiweb/rsdroid/BackendMutex.java
@@ -178,6 +178,16 @@ public class BackendMutex implements BackendV1 {
         }
     }
 
+    @Override
+    public void setPageSize(long pageSizeBytes) {
+        try {
+            lock.lock();
+            backend.setPageSize(pageSizeBytes);
+        } finally {
+            lock.unlock();
+        }
+    }
+
     // RustBackend Implementation
 
     @Override
--- a/rsdroid/src/main/java/net/ankiweb/rsdroid/BackendMutex.java
+++ b/rsdroid/src/main/java/net/ankiweb/rsdroid/BackendMutex.java
@@ -139,10 +139,10 @@ public class BackendMutex implements BackendV1 {
     }
 
     @Override
-    public Sqlite.DBResponse getPage(int page, int sequenceNumber) {
+    public Sqlite.DBResponse getNextSlice(long startIndex, int sequenceNumber) {
         try {
             lock.lock();
-            return backend.getPage(page, sequenceNumber);
+            return backend.getNextSlice(startIndex, sequenceNumber);
         } finally {
             lock.unlock();
         }
--- a/rsdroid/src/main/java/net/ankiweb/rsdroid/BackendV1Impl.java
+++ b/rsdroid/src/main/java/net/ankiweb/rsdroid/BackendV1Impl.java
@@ -18,6 +18,7 @@ package net.ankiweb.rsdroid;
 
 import androidx.annotation.CheckResult;
 import androidx.annotation.Nullable;
+import androidx.annotation.VisibleForTesting;
 
 import com.google.protobuf.InvalidProtocolBufferException;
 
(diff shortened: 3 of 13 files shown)


rsdroid/src/main/java/net/ankiweb/rsdroid/BackendMutex.java
```diff
@@ -139,10 +139,10 @@ public String getPath() {
     }
 
     @Override
-    public Sqlite.DBResponse getPage(int page, int sequenceNumber) {
+    public Sqlite.DBResponse getNextSlice(long startIndex, int sequenceNumber) {
```
mikehardy, line 142: `Page` did not have this effect but now with `Slice` the code review is making me hungry
in seriousness - between "chunk" or this, I like slice, nice

mikehardy (approved):
I'm definitely not qualified to review Rust after looking at it for a moment, and don't have the time to get there but the Java bits looked sane (although if you step back for a moment: it's kind of insanely complicated for a flashcard app! until you realize everyone wants multi-device syncing online backup and fast performance from 10,000 cards with 100,000 reviews and about 4 new features a week and then, yep - looks good!)

rslib-bridge/src/dbcommand.rs
```diff
@@ -12,7 +12,71 @@ use anki::backend_proto::{DbResponse, DbResult};
...
+    fn estimate_size(&self) -> usize {
+        self.fields.iter().map(|x| x.estimate_size()).sum()
+    }
+}
+
+impl Sizable for DbResult {
+    fn estimate_size(&self) -> usize {
+        // Performance: It might be best to take the first x rows and determine the data types
+        // If we have floats or longs, they'll be a fixed size (excluding nulls) and should speed
+        // up the calculation as we'll only
```
david-allison-1, line 56: TODO

rslib-bridge/src/dbcommand.rs
```diff
@@ -113,4 +218,136 @@ static mut SEQUENCE_NUMBER: i32 = 0;
 pub(crate) unsafe fn next_sequence_number() -> i32 {
     SEQUENCE_NUMBER = SEQUENCE_NUMBER + 1;
     SEQUENCE_NUMBER
+}
+
+/** The maximum size of a page returned from the database **/
+// static mut DB_COMMAND_PAGE_SIZE : usize = 1024 * 1024 * 4;
```
david-allison-1, line 224: Remove

rslib-bridge/src/dbcommand.rs
```diff
@@ -113,4 +218,136 @@ static mut SEQUENCE_NUMBER: i32 = 0;
...
+}
+
+fn get_max_page_size() -> usize {
+    *DB_COMMAND_PAGE_SIZE.lock().unwrap()
+}
+
+
+#[cfg(test)]
+mod tests {
+    // Note this useful idiom: importing names from outer (for mod tests) scope.
```
david-allison-1, line 243: Remove

rslib-bridge/src/dbcommand.rs
```diff
@@ -12,7 +12,71 @@ use anki::backend_proto::{DbResponse, DbResult};
...
+fn select_slice_of_size<'a>(mut rows : impl Iterator<Item = &'a Row>, max_size: usize) -> FoldWhile<(usize, Vec<Row>)> {
+    let init: Vec<Row> = Vec::new();
+    let folded = rows.fold_while((0, init), |mut acc, x| {
+        let new_size = acc.0 + x.estimate_size();
+        // If the accumulator is 0, but we're over the size: return a single result so we don't loop forever.
+        // Theoretically, this shouldn't happen as data should be reasonably sized
+        if new_size > max_size && acc.0 > 0 {
+            Done(acc)
+        } else {
+            acc.1.push(x.to_owned());
```
david-allison-1, line 74: Perf note: I'd expect it to be much faster to return (size, numElements) then do a bulk copy/slice later

Arthur-Milchior:
> The calculation for the data is inexact - string length, and 8 bytes
for doubles/longs.

Not clear to me. Do you mean that you are choosing to make an approximation here. That you are correcting this bug. That it can't be better

Arthur-Milchior:
> although if you step back for a moment: it's kind of insanely complicated for a flashcard app! until you realize everyone wants multi-device syncing online backup and fast performance from 10,000 cards with 100,000 reviews and about 4 new features a week and then, yep - looks good!

Actually, I believe the number of cards is really not the problem here. It's the fact that deck and note type lists are saved as a big json. Each data is small enough otherwise

Arthur-Milchior:
I don't know what page/slices are, what it means that it changes.

what limit the size. The part between Rust and Java, or the sqlite access?

Arthur-Milchior (commented):
Also hard to review. I don't know what should be done with DB.  I can read the rust but don't know what constraints must be satisfied.

Do you think it can eventually be reused and shared as an external library, seems a lot of work and I can imagine other people want to use rust in android as backend

rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/DatabaseIntegrationTests.java
```diff
@@ -38,6 +37,8 @@
 @RunWith(Parameterized.class)
 public class DatabaseIntegrationTests extends DatabaseComparison {
 
+    // determined by debugger
+    public static int DB_PAGE_NUM_INT_ELEMENTS = 111; // (8 + 1) * 111 is last number under 1000
```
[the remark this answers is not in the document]
david-allison-1: I'll clarify in the code.

1000 is the max page size

Rust handles the `INTEGER` type affinity as a long:
 * 8 bytes in memory
 * an additional byte for the overhead of an `optional`.

So floor(1000/(8 + 1))

david-allison-1:
> > The calculation for the data is inexact - string length, and 8 bytes
> > for doubles/longs.
> 
> Not clear to me. Do you mean that you are choosing to make an approximation here. That you are correcting this bug. That it can't be better

Yes, I'm choosing an approximation: realistically it's not worth the time to get an exact size of the protobuf. 

A bad implementation would be ridiculously slow, and a good implementation would likely be far too complicated to be worth the engineering effort. 

We also estimate the Rust in-memory size. With all these estimations, we were <10% off the 'proper' value, 2MB is a conservative number anyway, I'd expect that we could double it and not need to worry, so none of the estimations bother me.

> Actually, I believe the number of cards is really not the problem here. It's the fact that deck and note type lists are saved as a big json. Each data is small enough otherwise

Neither decks nor note types matter, as we should handle them via `loadColumn` and chunking based on the window size.

This is for the few queries where saying "1000 rows" could blow up and cause an OOM. This mostly occurs on the notes table - the `flds` column is unlimited length and a few queries iterate over this.

> I don't know what page/slices are, what it means that it changes.
> what limit the size. The part between Rust and Java, or the sqlite access?

Native code (Rust) has an effectively unlimited heap on Android. The JVM is severely limited.

I was advised to perform the same operation that the Python does with rslib: avoid streaming the results from queries from the database and pull everything into memory in one operation. Streaming is problematic due to locking on the collection object. We might need to revisit this in future, but to be honest, I doubt it.

So: we pull in the entire resultset into the Rust, and stream to the Java Cursor in 2MB "slices" to the Java, to avoid the JVM's heap limit.

mikehardy:
> > although if you step back for a moment: it's kind of insanely complicated for a flashcard app! until you realize everyone wants multi-device syncing online backup and fast performance from 10,000 cards with 100,000 reviews and about 4 new features a week and then, yep - looks good!
> 
> Actually, I believe the number of cards is really not the problem here. It's the fact that deck and note type lists are saved as a big json. Each data is small enough otherwise

I was stepping even farther. Why is a flashcards program rendering HTML/CSS/JS (with plugins!), based out of Java, going across a Protobuf bridge to Rust, where it does database access etc. It's sort of like a Rube Goldberg machine. But it all seems justifiable, just so many layers...

dae:
Well, to be fair, you had database cursors in Java calling out to a C++ wrapper over JNI which forwarded the requests on to the C SQLite bindings - most of the layers were already there, they were just someone else's responsibility when you were using requery :-)

All this complexity surrounding the database cursors is unfortunate, and it's considerably more complicated than the computer version's DB layer. But the low heap limit on Java means you're working with constraints that the computer version doesn't have.

The good news is that these changes are a stop-gap, and could potentially be made simpler in the future. Once AnkiDroid has been updated to call backend methods to mutate notes in bulk, it will no longer be necessary to divide large queries into chunks and go through all the extra ceremony of passing them to Java and back - the backend can just make the changes directly. Most of the backend functionality is already implemented; the one remaining task that springs to mind is the import/export code.

mikehardy:
I appreciate that you actually responded to my Rube-Goldberging of the system @dae haha - for the avoidance of doubt I honestly do believe each layer is suiting it's purpose well. To take it farther in the humor direction, I hope that someone implements a card template in WASM using our javascript methods, and I will personally send them a dollar if their original language before compiling to WASM was...Rust :)

rsdroid/src/main/java/net/ankiweb/rsdroid/BackendV1Impl.java
```diff
@@ -340,6 +341,16 @@ private void performTransaction(String kind) {
         }
     }
 
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    public static void setPageSizeForTesting(long pageSizeInBytes) {
+        // TODO: Make this nonstatic
```
david-allison-1, line 346: remove

Issue #46, opened 2021-03-27
RustBackendLoader: "Can't find dependent libraries"


david-allison-1:
https://github.com/ankidroid/Anki-Android/pull/8345#issuecomment-808678996

> The error I am getting 
![image](https://user-images.githubusercontent.com/46752548/112713503-7bce8680-8efb-11eb-943f-3bbed39c8b48.png)

david-allison-1:
Trimmed dumpbin output

```
C:\Users\David\Downloads>"C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Tools\MSVC\14.16.27023\bin\Hostx64\x64\dumpbin" /dependents rsdroid1625960893423553494.dll
Dump of file rsdroid1625960893423553494.dll

File Type: DLL

  Image has the following dependencies:

    ole32.dll
    SHELL32.dll
    ADVAPI32.dll
    libgcc_s_seh-1.dll
    KERNEL32.dll
    msvcrt.dll
    USERENV.dll
    WS2_32.dll
C:\Users\David\Downloads>
```

david-allison-1:
User comment:

> I checked my temp directory it doesn't exist.

david-allison-1:
For testing:

```
david@Auxiliary:/mnt/c/Users/David/Downloads/Anki/rsd$ objdump --private-headers rsdroid1625960893423553494.dll  | grep "DLL Name:"
        DLL Name: ole32.dll
        DLL Name: SHELL32.dll
        DLL Name: ADVAPI32.dll
        DLL Name: libgcc_s_seh-1.dll
        DLL Name: KERNEL32.dll
        DLL Name: msvcrt.dll
        DLL Name: USERENV.dll
```

david-allison-1:
Was suggested to try `-Ctarget-feature=+crt-static`

david-allison-1:
Failing test: https://github.com/david-allison-1/Anki-Android-Backend/runs/2272403365?check_suite_focus=true

Starkly:
Just to note, libgcc_s_seh-1.dll has a dependency on libwinpthread-1.dll

if both are in your PATH, then the library will load and the tests should run ok, without installing all of gcc

mikehardy:
Everyone should use a package manager, and on windows I use chocolatey, install it like so https://chocolatey.org/install#install-step2

Then:

`choco install mingw`

[...]

PR #72, opened 2021-04-05
rsdroid-testing: Diagnose missing Windows lib

david-allison:
The rsdroid dll is currently dynamically linked to gcc to allow unwinding. I was unable to find documentation to disable this.

This isn't ideal for Windows consumers, as it requires that they install GCC (specifically: `libgcc_s_seh-1.dll`).

So we inform the user that they need to install GCC - not ideal, but it fixes a blocker

Workaround for issue #46


dae:
Maybe another option would be to include the gcc dll next to the rsdroid one in the package you distribute?

PR #75, opened 2021-04-10
Allow multiple usages of RustCleanup annotation

david-allison-1:
Also defines it as a source-only annotation

### DIFF
--- a/rsdroid/src/main/java/net/ankiweb/rsdroid/RustCleanupCollection.java
+++ b/rsdroid/src/main/java/net/ankiweb/rsdroid/RustCleanupCollection.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 David Allison <davidallisongithub@gmail.com>
+ * Copyright (c) 2021 David Allison <davidallisongithub@gmail.com>
  *
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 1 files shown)


rsdroid/src/main/java/net/ankiweb/rsdroid/RustCleanupCollection.java
```diff
@@ -0,0 +1,25 @@
...
+ * this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package net.ankiweb.rsdroid;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.SOURCE)
+public @interface RustCleanupCollection {
```
mikehardy, line 23: One thing that might help me with the annotations is if there was some small javadoc on the interface that described under what conditions you would mark something with it, and what the intended change is for the future. If I saw "RustCleanupCollection" somewhere I would know something needed to be done but not exactly what, and I wouldn't know (after seeing it) if (as I'm reading other stuff) it should be applied there too but maybe just didn't exist yet

david-allison-1: Thanks for keeping me honest. I'll add it to the source and commit message

This annotation isn't actually to be used, it's an endorsed Java-level convention to allow for multiple annotations on a single class. So a user can type `@RustCleanup("foo") @RustCleanup("bar")`

PR #77, opened 2021-04-18
fix[CI]: update Lexical core

david-allison-1:
0.7.4 is incompatible with Rust nightlies

```
error[E0277]: cannot subtract `usize` from `u32`
    --> lexical-core-0.7.4/src/atof/algorithm/math.rs:2065:25
     |
2065 |     let rs = Limb::BITS - s;
     |                         ^ no implementation for `u32 - usize`
     |
     = help: the trait `Sub<usize>` is not implemented for `u32`
 ```

### DIFF
--- a/rslib-bridge/Cargo.toml
+++ b/rslib-bridge/Cargo.toml
@@ -19,6 +19,7 @@ serde_derive = "1.0.114"
 lazy_static = "1.4.0"
 num_enum = "0.5.0"
 itertools = "0.10.0"
+lexical-core = "0.7.5"
 
 # picked bundled - TODO: Is this correct?
 rusqlite = { version = "0.23.1", features = ["trace", "functions", "collation", "bundled"] }


mikehardy:
Is there an "open source repository maintainer" equivalent to the idea of the "hedonic treadmill"?

Like, you are trying to improve software things ("do pleasurable things" in the metaphor) but to do so involves more and more yak shaves until inevitably you are maintaining all the software in the universe in order to improve things even a slight bit more? (which is like metaphorically falling prey to opiate addiction or something).

Is there a solution to this? asking for a friend

:lotus_position:

dae:
Heh :-) Certainly know how that feels.

@david-allison-1 re the use of nightly, output products are available in a fairly predictable location even in stable builds - you might be able to avoid nightly altogether, and just copy the required files out of the build folder at the end of the build.

david-allison-1:
Definitely feasible! We'll get it in along with CI improvements in https://github.com/david-allison-1/Anki-Android-Backend/pull/39. Rust almost certainly can't change the directory structure due to other projects depending on it, and the flag would just be to avoid a copy.

Code fef: 
https://github.com/david-allison-1/Anki-Android-Backend/pull/39/files#diff-8121a0ad08d5b865324560c00398477f4ed33a65682817684b4da0d29c6290c4R155-R179

PR #79, opened 2021-04-23
Add downgradeDatabase method

david-allison-1:
* open collection (don't upgrade, do create, fail if not v16)
* close collection (downgrade = true)

We need V16 exactly as the `close` method doesn't handle closing from anything other than the latest schema (crashes on downgrading a V11 database to v11) - might work on v15, but we don't expect a V15 database

Upgrades `anki` submodule to make `.close()` public

Related: https://github.com/ankidroid/Anki-Android/issues/8529


rsdroid/src/main/java/net/ankiweb/rsdroid/AnkiDroidBackendImpl.java
```diff
@@ -34,6 +34,16 @@ public Pointer ensureBackend() {
         return NativeMethods.executeAnkiDroidCommand(backendPointer, command, args);
     }
 
+    public void downgradeBackend(String collectionPath) {
+        String ret = NativeMethods.downgradeDatabase(collectionPath);
```
mikehardy, line 38: I always chuckle when I see this. In the react-native world, when you are in the javascript level (majority of the time coding app stuff and libraries even) and you are writing some java code we uniformly consider that "native" and refer to the Java module as the NativeModule.

It's turtles all the way down. I wonder if in the Rust code there is some C or assembly and rust calls it NativeMethods

mikehardy (approved):
As much as I know Rust (not much) LGTM

PR #88, opened 2021-05-07
Remove `protobuf-compiler` pip pakacge

TarekkMA:
fixes: #88

### DIFF
--- a/.github/workflows/linux_build.yml
+++ b/.github/workflows/linux_build.yml
@@ -34,8 +34,10 @@ jobs:
     - name: Install Rust Targets
       run: .github/scripts/install_rust_targets.sh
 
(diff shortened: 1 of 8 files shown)


TarekkMA:
@david-allison-1 The checks are green. should I also remove the pyenv/protobuf scripts, since they are not used anymore?

TarekkMA:
This PR will shave about 10 minutes in CI. A step into the right direction 🎉

david-allison-1 (approved):
LGTM

david-allison-1:
> @david-allison-1 The checks are green. should I also remove the pyenv/protobuf scripts, since they are not used anymore?

Sure! I'll add a follow-up isue

PR #89, opened 2021-05-08
Add rust build cache

### LINKED ISSUE #10: Enable dependency caching for Rust Build
If we cache `/target/`, we should massively be able to cut down on the build times.

This may be problematic on macOS. See: https://github.com/ankitects/anki/commit/43f5db648078e9cd4e18ca2559eae396858327d1

TarekkMA:
Based on #88 
Closes #10.

I didn't include cache in the publishing workflows. should I add it or leave it as it's.

### DIFF
--- a/.github/workflows/linux_build.yml
+++ b/.github/workflows/linux_build.yml
@@ -46,7 +46,7 @@ jobs:
           ~/.cargo/registry
           ~/.cargo/git
           rslib-bridge/target
-        key: ${{ runner.os }}-cargo-${{ hashFiles('rslib-bridge/**/Cargo.lock') }}
+        key: ${{ runner.os }}-assembleRelease-rust-${{ hashFiles('rslib-bridge/**/Cargo.lock') }}
 
     - name: Build
       run: ./gradlew clean assembleRelease -DtestBuildType=release -Dorg.gradle.daemon=false -Dorg.gradle.console=plain # assembleAndroidTest
--- a/.github/workflows/linux_build.yml
+++ b/.github/workflows/linux_build.yml
@@ -39,6 +39,15 @@ jobs:
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
 
+    - name: Rust Cache
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+          rslib-bridge/target
+        key: ${{ runner.os }}-cargo-${{ hashFiles('rslib-bridge/**/Cargo.lock') }}
+
     - name: Build
       run: ./gradlew clean assembleRelease -DtestBuildType=release -Dorg.gradle.daemon=false -Dorg.gradle.console=plain # assembleAndroidTest
 
--- a/.github/workflows/linux_build.yml
+++ b/.github/workflows/linux_build.yml
@@ -46,7 +46,10 @@ jobs:
           ~/.cargo/registry
           ~/.cargo/git
           rslib-bridge/target
-        key: ${{ runner.os }}-assembleRelease-rust-${{ hashFiles('rslib-bridge/**/Cargo.lock') }}
+        key: ${{ runner.os }}-rust-v1-assembleRelease-${{ hashFiles('rslib-bridge/**/Cargo.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-rust-v1-assembleRelease
+          ${{ runner.os }}-rust-v1
 
     - name: Build
       run: ./gradlew clean assembleRelease -DtestBuildType=release -Dorg.gradle.daemon=false -Dorg.gradle.console=plain # assembleAndroidTest
--- a/.github/workflows/linux_build.yml
+++ b/.github/workflows/linux_build.yml
@@ -34,8 +34,10 @@ jobs:
     - name: Install Rust Targets
       run: .github/scripts/install_rust_targets.sh
 
-    - name: Install google protobuf compiler
-      run: .github/scripts/linux_install_protobuf.sh
+    - name: Install Protoc
+      uses: arduino/setup-protoc@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
 
     - name: Build
       run: ./gradlew clean assembleRelease -DtestBuildType=release -Dorg.gradle.daemon=false -Dorg.gradle.console=plain # assembleAndroidTest
--- a/.github/workflows/macos_build.yml
+++ b/.github/workflows/macos_build.yml
@@ -50,7 +50,10 @@ jobs:
           ~/.cargo/registry
           ~/.cargo/git
           rslib-bridge/target
-        key: ${{ runner.os }}-assembleDebug-rust--${{ hashFiles('rslib-bridge/**/Cargo.lock') }}
+        key: ${{ runner.os }}-rust-v1-assembleDebug-${{ hashFiles('rslib-bridge/**/Cargo.lock') }}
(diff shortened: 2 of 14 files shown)


david-allison-1:
Let's leave publishing

TarekkMA:
# Old Times
Screen Shot 2021-05-08 at 2 54 57 AM
Screen Shot 2021-05-08 at 2 55 09 AM

TarekkMA:
Results are promising, this PR along with protobuff-compiler one will make CI finish in about 10 mins

Still I need to put workflow ID or something in the cache key, as rsdroid macos cache seems to be the same as Robolectric one.

TarekkMA:
I have rebased over #88 so i can see the full effect both PR will have on timing, also made the cache key dependant on the Gradle task that will run

TarekkMA:
# With google protoc 
Screen Shot 2021-05-08 at 2 59 48 PM
Screen Shot 2021-05-08 at 2 59 58 PM

TarekkMA:
# Both google protoc and rust caching
Screen Shot 2021-05-08 at 3 23 05 PM
Screen Shot 2021-05-08 at 3 23 20 PM

.github/workflows/linux_build.yml
```diff
@@ -34,8 +34,19 @@ jobs:
...
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Rust Cache
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+          rslib-bridge/target
+        key: ${{ runner.os }}-assembleRelease-rust-${{ hashFiles('rslib-bridge/**/Cargo.lock') }}
```
mikehardy, line 49: Having just done a lot of work on caching in another repo, you might want a setup like this:

```suggestion
        key: ${{ runner.os }}-assembleRelease-rust-v1-${{ hashFiles('rslib-bridge/**/Cargo.lock') }}
        restore-keys: ${{ runner.os }}-assembleRelease-rust-v1
```

- The `v1` is because GitHub has no way of clearing caches ever. They are immutable. So you use a "primary key" which is unique for whatever your software state is (Cargo.lock is a nice choice for rust) but then just in case you need to drop the cache, you include a cache version specifier in a slightly less unique path component, and you can increment to v2 if needed. The other cache drop solution is to actually define a repo secret and include it as `{{ secrets.CACHE_VERSION }}` or similar if you want to avoid having to do a commit to purge a corrupt cache
- the `restore-keys` lets you at least get a partial cache hit (outdated, but it will save build time still) in case Cargo.lock changes, but since it is not the primary key a successful build from the restore-keys cache (but not the primary key) will upload a new cache for future builds, so you get some acceleration and still update, it's nice https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key

I would apply that style to all of them that use caching

TarekkMA: Thank you a lot for the feedback.

I thought about what will happen if `hashFiles('rslib-bridge/**/Cargo.lock')` and I was not very comfortable with the CI running full 45+ minutes to do a full build again. but the `restore-keys` is a very good solution.

TarekkMA: I will make the cache follow this naming schema: 
```
${{ runner.os }}-${{ cache_name }}-${{  cache_version }}-${{ task }}-${{ hash }}
```
for example, the above snippet will be: 
```yml
    - name: Rust Cache
      uses: actions/cache@v2
      with:
        path: |
          ~/.cargo/registry
          ~/.cargo/git
          rslib-bridge/target
        key: ${{ runner.os }}-rust-v1-assembleRelease-${{ hashFiles('rslib-bridge/**/Cargo.lock') }}
        restore-keys: |
          ${{ runner.os }}-rust-v1-assembleRelease
          ${{ runner.os }}-rust-v1
```

TarekkMA:
The workflows are getting bigger and more complex and have lots of duplicates. I don't know how this issue should be solved

.github/workflows/linux_build.yml
mikehardy: yeah - that seems like it follows what I know, from the key matching documentation and watching waaaaay way way too much build runs in the last week :-)

mikehardy:
The issue of duplicate workflows is one that has really been bugging me as well. I have the same issue to the next power - I have a lot of complicated workflows I'm maintaining now that are duplicated as a set in a repo, but are then duplicated across a bunch of repos. LIke 6 at this point, with 2 more planned so I am really struggling.

So far the official answer is composite run step actions where you make an action that can be a series of "run" (that is, shell script) steps but you can't call an action from within an action https://stackoverflow.com/questions/58860769/call-a-github-action-from-another-github-action#58869243

It is possible to have a "local action" - when I built buildcache-action I realized that's actually how you test an action: https://github.com/mikehardy/buildcache-action/blob/d093164b5e9ef9aed92497e161a226f3dd1910f7/.github/workflows/react-native-build-test.yml#L28

So, you could think in a "meta" / higher level sense about what parts are duplicate, how you would do the equivalent of an "extract method" refactor on them by figuring out the common inputs / outputs, then make that a reusable chunk as a local action.

Others use sub-scripts in a scripts directory to share logic, like the FlutterFire project https://github.com/FirebaseExtended/flutterfire/tree/master/.github/workflows/scripts - If you're good with a cross-platform scripting language like javascript or similar. I suppose it's possible to keep it cross-platform and reusable.

I don't have a great solution though - every action in the workflow brings a maintenance burden of upgrades and semantic versioning etc and potential for bugs and there's no good way to have a central one across multiple repos that is itself version controlled etc.

If you find a good solution let me know please

TarekkMA:
Thank you for sharing this, I was looking at composite actions yesterday and I was disappointed to know that we can't run actions inside them.  

Other CI providers like Azure pipelines provide a feature called templates, this is exactly what we are looking for. but unfortunately, GitHub actions don't have this yet. 😞

david-allison-1:
@TarekkMA Was this good to go? Apologies that I missed the follow-ups

TarekkMA:
> @TarekkMA Was this good to go? Apologies that I missed the follow-ups

No worries, yes I think this is good to go.

Note that this PR is based on #88

david-allison-1:
@TarekkMA Needs a de-conflict and good to go, thanks so much for this!!!

mikehardy:
> Thank you for sharing this, I was looking at composite actions yesterday and I was disappointed to know that we can't run actions inside them.
> 
> Other CI providers like Azure pipelines provide a feature called templates, this is exactly what we are looking for. but unfortunately, GitHub actions don't have this yet. disappointed

I'm still thinking about this
The only solution I can think of is to package things up as Actions still, perhaps with some sort of plugin design where if you add things to .github/<somedir> it will iterate and do them, sort of like database migration systems take care of a lot of infrastructure / boilerplate, then just iterate over an ordered set of scripts that use a common API / set of hooks that would be provided by / called by the executor engine in the action. Still thinking about this, it is a huge pain point for me

mikehardy:
I fixed the conflicts using the GitHub UI. Yes, that puts a merge commit in there but it was going to be a squash anyway based on prior commit history

david-allison-1 (approved):
Looks fantastic! My fault for not getting to this sooner

david-allison-1:
ugh.... rebased rather than squashed

mikehardy:
hahaha - I just did that in Anki-Android. Sigh. At least the merge commit was small. Condolences. You could probably force push and no one would care since no one else has pulled it.

PR #100, opened 2021-05-25
fix ContentProvider crash

david-allison-1:
The internals of ContentProvider call (un)registerContentObserver

We don't need to do anything here aside from not crash, as `requery()` is not called, so we don't need to notify of any changes

Fixes API-related crashes in AnkiDroid


david-allison-1:
@mikehardy I'm not familiar with this functionality of ContentProviders, and wasn't aware that it was used internally.

Could you give this a quick look over and confirm that it's OK with your knowledge.

This fixes Akebi on my device (tested locally)

mikehardy (approved):
I'll trust you on that one - I don't know ContentProviders to this depth either, but you reproduced it
If you want to build this and make a PR I can get it out in 2.15.1 - I can't see how it could harm really

PR #104, opened 2021-05-26
DatabaseCursor: Implement ContentProvider methods

david-allison-1:
Our ContentProvider/API implementation used:

`android.database.CursorToBulkCursorAdaptor`
and `DatabaseUtils.cursorFillWindow`

These called through to `moveToPosition` and `getType`
which were not implemented.

Fix this by implementing the methods


mikehardy (approved):
Seems straightforward enough?
0-based array index usage seems consistent

PR #105, opened 2021-05-26
remove legacy DB implementations

david-allison-1:
Removed in favor of StreamingProtobufSQLiteCursor


mikehardy (approved):
Going to trust you on that one, but I like the look of +0 / -427 if everything still works

PR #108, opened 2021-05-26
DB Cursors: allow moving backwards 

david-allison-1:
We should support this, as resetting the cursor is required by `DatabaseUtils.cursorFillWindow`, which triggered issues with API clients

Note: Inefficient on `moveToPrevious`, but I don't see this as a common case

### DIFF
--- a/rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/DatabaseIntegrationTests.java
+++ b/rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/DatabaseIntegrationTests.java
@@ -33,6 +33,7 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.isOneOf;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 @RunWith(Parameterized.class)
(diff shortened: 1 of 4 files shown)


rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/DatabaseIntegrationTests.java
```diff
@@ -286,6 +287,46 @@ public void testRowCountPageAndOne() {
...
+        for (int i = 0; i < DB_PAGE_NUM_INT_ELEMENTS + 1; i++) {
+            db.execSQL("insert into test VALUES (1)");
+        }
+
+        try (Cursor c = db.query("select * from test")) {
+            assertThat(c.getPosition(), is(-1));
+            assertTrue(c.moveToFirst());
+            assertThat(c.getPosition(), is(0));
+            // despite returning false, it works
+            assertFalse("moveToPosition(-1) should return false, but should work", c.moveToPosition(-1));
```
mikehardy, line 325: that's crazy

rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/database/StreamingProtobufSQLiteCursorTest.java
```diff
@@ -73,6 +135,18 @@ private void iterateAllRows(SupportSQLiteDatabase db) {
         }
     }
 
+    private void iterateAllRowsBackwards(SupportSQLiteDatabase db) {
```
mikehardy, line 138: This is really nitpicky, but there is an assumption that you're handed a database where value @ col 0 @ row N == N, maybe `checkValueAtRowEqualsRowNumBackwards`. Or on second thought ignore me

rsdroid/src/main/java/net/ankiweb/rsdroid/database/StreamingProtobufSQLiteCursor.java
```diff
@@ -85,25 +86,21 @@ public int getCount() {
 
     @Override
     public int getPosition() {
```
mikehardy, line 88: interesting - I see the type switches around here, I'm guessing the Cursor interface returns int, so we have to use it? I suppose we won't ever need a `long` worth of positions?

rsdroid/src/main/java/net/ankiweb/rsdroid/database/StreamingProtobufSQLiteCursor.java
```diff
@@ -65,11 +66,11 @@ public StreamingProtobufSQLiteCursor(SQLHandler backend, String query, Object[]
...
 
-    private void getNextPage() {
-        position = -1;
-
+    private void loadPage(long startingAtIndex) {
         try {
-            results = backend.getNextSlice(getSliceStartIndex() + getCurrentSliceRowCount(), sequenceNumber);
+            long requestedIndex = startingAtIndex == -1 ? 0 : startingAtIndex;
+            results = backend.getNextSlice(requestedIndex, sequenceNumber);
+            position = startingAtIndex == -1 ? -1 : 0;
```
mikehardy, line 73: is this right? The position is always -1 or 0, even if startingAtIndex was deeper into the result set? so it's more like positionInCurrentPage vs position-in-whole-result-set? Or am I totally misunderstanding this?

mikehardy: Ah I see positionGlobal later - maybe positionSlice to disambiguate

mikehardy (approved):
Nothing looks off - just nitpicks and/or questions, approving in general and happy to see it go in as is if none of my stuff seems worthwhile (and it may not)

rsdroid/src/main/java/net/ankiweb/rsdroid/database/StreamingProtobufSQLiteCursor.java
david-allison-1: Rust returns longs, the Cursor interface returns ints

rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/DatabaseIntegrationTests.java
david-allison-1: Very happy I spent the time comparing these tests between the Rust and the Framework... saved quite a few mishaps here

PR #111, opened 2021-06-02
Fix: db string conversion

david-allison-1:
loading from TEXT -> Long required strtoll
loading from TEXT -> Double required strtod

Both of these are copied implementations

Related: https://github.com/ankidroid/Anki-Android/issues/8877

### DIFF
--- a/rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/DatabaseIntegrationTests.java
+++ b/rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/DatabaseIntegrationTests.java
@@ -204,6 +204,27 @@ public class DatabaseIntegrationTests extends DatabaseComparison {
     }
 
     @Test
+    public void testLongConversions() {
+        testLongConversion(SQLOutput.asInt("1"), 1L);
+        testLongConversion(SQLOutput.asFloat("1.6"), 1L);
+        testLongConversion(SQLOutput.asNull(), 0L);
+        testLongConversion(SQLOutput.asText("hi"), 0L);
+        testLongConversion(SQLOutput.asText("2"), 2L);
+        testLongConversion(SQLOutput.asText("2.52"), 2L);
+        testLongConversion(SQLOutput.asInt("9223372036854775808"), Long.MAX_VALUE);
+    }
+
+    @Test
+    public void testFailingLongConversions() {
+        try {
+            testLongConversion(SQLOutput.asBlob(), 42L);
+        } catch (SQLiteException e) {
+            assertThat(e.getMessage(), isOneOf("unknown error (code 0): Unable to convert BLOB to long",
+                    "unknown error (code 0 SQLITE_OK): Unable to convert BLOB to long"));
+        }
+    }
+
+    @Test
     public void testFloatConversions() {
         testFloatConversion(SQLOutput.asInt("1"), 1.0f);
         testFloatConversion(SQLOutput.asFloat("1.6"), 1.6f);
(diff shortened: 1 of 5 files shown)


rsdroid/src/main/java/net/ankiweb/rsdroid/utils/StringToLong.java
```diff
@@ -0,0 +1,81 @@
...
+        if (start == pos) {
+            return 0;
+        }
+
+        s = s.substring(start, pos);
+
+        try {
+            return Long.parseLong(s, radix);
+        } catch (NumberFormatException e) {
+            return Long.MAX_VALUE;
```
david-allison-1, line 78: check

rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/DatabaseIntegrationTests.java
```diff
@@ -186,6 +188,8 @@ public void testIntConversions() {
         testIntConversion("double", 1); // this is a .floor
         testIntConversion("string", 0);
         testIntConversion("null", 0);
+        testIntFromStringConversion("2", 2);
+        testIntFromStringConversion("2.52", 2);
```
mikehardy, line 192: In all of these, what about the error cases? Testing conversion from "IAmNotAnInt" to int / float / double etc, or just empty space, or null ? :thinking:

david-allison-1: This is testing the error cases: 

https://github.com/david-allison-1/Anki-Android-Backend/blob/626d65076534653e4f3334c98070c12b49c4705d/rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/DatabaseIntegrationTests.java#L421-L430

rsdroid/src/main/java/net/ankiweb/rsdroid/database/StreamingProtobufSQLiteCursor.java
```diff
@@ -269,5 +269,22 @@ public int getType(int columnIndex) {
...
+        } catch (NumberFormatException exception) {
+            return 0;
+        }
+    }
+
+    private double strtod(String stringValue) {
+        try{
+
+        return StringToDouble.strtod(stringValue);
+        } catch (NumberFormatException exception) {
```
mikehardy, line 285: spacing is off here

mikehardy (approved):
2 tiny things but nothing really substantive --> :+1:

rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/DatabaseIntegrationTests.java
david-allison-1: I'll get this rewritten when I have time tonight - it's minor on the TODO list, but let's fix it now it's been flagged

dae:
One alternative that might be worth considering is casting the strftime output at the sql level. This would make the conversion more explicit, avoid the need for pasted-in code, and prevent accidental usage of a string when a number was expected.

```
sqlite> select cast("1.6" as int);
cast("1.6" as int) = 1
sqlite> select cast("1.6" as float);
cast("1.6" as float) = 1.6
```

mikehardy:
sqlite should run that stuff a great deal faster as well - good thought

david-allison-1:
Added a separate issue

Avoiding the need for this conversion via SQL is the ideal, but we should honor the contract specified in `CursorWindow` regardless: 

> If the field is of type Cursor#FIELD_TYPE_STRING, then the result is the value obtained by parsing the string value with strtoll.

https://developer.android.com/reference/android/database/CursorWindow#getLong(int,%20int)

mikehardy (approved):
Tests are a lot more readable now

PR #125, opened 2021-07-07
Add BackendVNextFactory and support classes

david-allison-1:
Allows for schema V16 handling

We don't hardcode the version in the class name, as that means we'll need to change later

Lots of boilerplate due to Android Db requirements, sadly.

### DIFF
--- a/rsdroid/src/main/java/net/ankiweb/rsdroid/BackendFactory.java
+++ b/rsdroid/src/main/java/net/ankiweb/rsdroid/BackendFactory.java
@@ -16,6 +16,10 @@
 
 package net.ankiweb.rsdroid;
 
+import androidx.sqlite.db.SupportSQLiteOpenHelper;
+
+import net.ankiweb.rsdroid.database.RustSQLiteOpenHelperFactory;
+
 public class BackendFactory {
 
     private BackendV1 backend;
@@ -49,5 +53,9 @@ public class BackendFactory {
 
         // we could swallow the exception here, most of the time it will be "collection is already closed"
         backend.closeCollection(false);
+    }
+    
+    public SupportSQLiteOpenHelper.Factory getSQLiteOpener() {
+        return new RustSQLiteOpenHelperFactory(this);
     }
 }
--- a/rsdroid/src/main/java/net/ankiweb/rsdroid/BackendFactory.java
+++ b/rsdroid/src/main/java/net/ankiweb/rsdroid/BackendFactory.java
@@ -18,9 +18,7 @@ package net.ankiweb.rsdroid;
 
 import androidx.sqlite.db.SupportSQLiteOpenHelper;
 
-import net.ankiweb.rsdroid.database.RustSQLiteOpenHelperFactory;
-
-public class BackendFactory {
+public abstract class BackendFactory {
 
     private BackendV1 backend;
 
@@ -29,14 +27,9 @@ public class BackendFactory {
 
     }
 
-    /**
-     * Obtains an instance of BackendFactory which will connect to rsdroid.
-     * Each call will generate a separate instance which can handle a new Anki collection
-     */
-    @RustV1Cleanup("RustBackendFailedException may be moved to a more appropriate location")
+    @RustCleanup("Use BackendV[11/Next]Factory")
     public static BackendFactory createInstance() throws RustBackendFailedException {
-        NativeMethods.ensureSetup();
-        return new BackendFactory();
+        return BackendV11Factory.createInstance();
     }
 
     public synchronized BackendV1 getBackend() {
@@ -55,7 +48,5 @@ public class BackendFactory {
         backend.closeCollection(false);
     }
     
-    public SupportSQLiteOpenHelper.Factory getSQLiteOpener() {
-        return new RustSQLiteOpenHelperFactory(this);
-    }
+    public abstract SupportSQLiteOpenHelper.Factory getSQLiteOpener();
 }
--- a/rsdroid/src/main/java/net/ankiweb/rsdroid/database/RustV11SupportSQLiteOpenHelper.java
+++ b/rsdroid/src/main/java/net/ankiweb/rsdroid/database/RustV11SupportSQLiteOpenHelper.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) 2020 David Allison <davidallisongithub@gmail.com>
+ *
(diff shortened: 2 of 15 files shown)


rsdroid/src/main/java/net/ankiweb/rsdroid/database/RustV11SupportSQLiteOpenHelper.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    protected SupportSQLiteDatabase createRustSupportSQLiteDatabase(@SuppressWarnings("SameParameterValue") boolean readOnly) {
+        Timber.d("createRustSupportSQLiteDatabase");
+        if (configuration != null) {
+            BackendV1 backend = backendFactory.getBackend();
+            BackendUtils.openAnkiDroidCollection(backend, configuration.name);
+            return new RustSupportSQLiteDatabase(backend, readOnly);
+        } else {
+            return new RustSupportSQLiteDatabase(backend, readOnly);
```
mikehardy, line 48: Not sure this is even a valid thought or not  and/or useful even if valid - but if the typing is changing from "open a database via rust" to "open a v11 database via rust", should there by any checks here for actual version compatibility, e.g. that it *is* actually a v11 (or compatible) database?

rsdroid/src/main/java/net/ankiweb/rsdroid/database/RustVNextSQLiteOpenHelperFactory.java
```diff
@@ -0,0 +1,39 @@
...
+
+import androidx.annotation.NonNull;
+import androidx.sqlite.db.SupportSQLiteOpenHelper;
+
+import net.ankiweb.rsdroid.BackendFactory;
+
+/**
+ * Implementation of {@link SupportSQLiteOpenHelper.Factory} using the Anki Desktop backend
+ */
+public class RustVNextSQLiteOpenHelperFactory implements SupportSQLiteOpenHelper.Factory {
```
mikehardy, line 27: So what's the plan with versioning then, thinking forward?
If we have V11 and VNext that implies that sometimes we need versioning pinned or it would be VPast and VNext.
If there is a future state where things become imcompatible, will VNext be peeled off to V42 (arbitrary number) with old-school implementation as necessary while VNext rumbles forward to stay "current"?

rsdroid/src/main/java/net/ankiweb/rsdroid/BackendFactory.java
```diff
@@ -29,14 +27,9 @@ protected BackendFactory() {
 
     }
 
-    /**
-     * Obtains an instance of BackendFactory which will connect to rsdroid.
-     * Each call will generate a separate instance which can handle a new Anki collection
-     */
-    @RustV1Cleanup("RustBackendFailedException may be moved to a more appropriate location")
+    @RustCleanup("Use BackendV[11/Next]Factory")
     public static BackendFactory createInstance() throws RustBackendFailedException {
```
mikehardy, line 31: From your commit message:
- I thought there would be a getInstance, it appears you meant createInstance?
- upstream/downstream confusion - if a change you can make affects others, isn't that like dropping pollution in a river and it flows...downstream? I always think of API consumers as downstream personally. But semantics is pedantic and not actually important :-)

mikehardy (commented):
Looks good in general, I mean, it certainly looks like functioning code, no reason to block.
I picked a couple nits and I am curious on future naming strategy since "VNext" looks like an open-ended dependency resolution like `+` which have their rare-but-important place but also should be double-checked prior to being put in use

rsdroid/src/main/java/net/ankiweb/rsdroid/database/RustV11SupportSQLiteOpenHelper.java
david-allison-1: This is intended to be (and is) checked in the application layer (Anki-Android). It's also checked in the rust:

https://github.com/david-allison-1/Anki-Android-Backend/blob/0030ebcdf0d32fcde2b8aad36a212385680b7836/rslib-bridge/src/sqlite.rs#L125-L129

rsdroid/src/main/java/net/ankiweb/rsdroid/database/RustVNextSQLiteOpenHelperFactory.java
david-allison-1: Versioning consists of 3 separate aspects. I presume the first upgrade will be a trial by fire, but my plan is as follows:

We have 3 concepts that we need to consider

* Schema
* ABI (methods/classes)
* ~~Translations~~ (not handled yet)

Moving to a new stable version of Anki can potentially change all 3, typically the changes will be the ABI/translations.

In this case, we can define "BackendV2", a Java interface of the new/changed functionality, code our Anki-Android implementations to both in our anti-corruption layer, and then switch over the implementation of the backend, which populates the V2 interface, and deprecate the `V1` only functionality

Schema changes are more involved:

* Aim to forward-port the "downgrade" logic from the new backend in a new minor version of AnkiDroid (we've done this for V16)
   *  This allows AnkiDroid consumers to automatically handle downgrades of the app as long as they don't make massive version advances
* Fix the ABI, as above

A schema change should then "just" be handled in the new code in `openCollection`

But, I foresee that this will take a while to get right, so while in development, allow:

* One AnkiDroid backend which uses the old ABI and the old database 
* One AnkiDroid backend, using the new ABI, with the new database (to enable testing, then move to release)

Only release one of these in release builds

If there are serious problems, then we'd peel off V42, and get that working as you propose. I've done it for v11 right now, but I don't expect that it'll be worth the engineering effort going forwards (better to track upstream, rather than spending time on temporary scaffolding).

mikehardy: Sounds excellent - looks like wiki fodder :-)

mikehardy (approved):
Looks like the updated commit message I nit-picked on isn't pushed yet but I assume it will be, still LGTM as is now other than that

david-allison-1:
I'll aim to get this into a release by Monday. Want to do a small amount of linking to `Anki-Android` to ensure that the `openCollection` change doesn't affect any live code (it shouldn't, but I'd rather not just leave this to automated tests).

PR #169, opened 2022-04-07
docs: clean up and update docs

mikehardy:
all docs: de-lint

env: used headings instead of checkboxes
env: fixed a link to rsdroid
env: added information where ever it was missing
env: noted specific version pins we need
env: added build command that may be used to test setup

I have successfully run the commands listed in `.github/workflows` via gradlew with the steps as documented in my updates!


docs/OVERVIEW.md
```diff
@@ -1,48 +1,52 @@
...
 See the components section for additional details
 
-### Setup
+### Command Setup
 
 ![](https://www.planttext.com/api/plantuml/img/ZLLDZzem4BtxLupsi0Af8FNAeThjqWE7TYMazj9MaUDCmKkmhV44YNzVEmci7Q8bkIJnl3VpvYjvyYo9csCjf69Bq5tPiuV68maNS5ff9mt3jl6y9gkhhr8Tq5GHD3mJfMrC9UaC3y_ce1VFfehMHMz-o1psXpzPrtxCyElpvkZgIvXBX1GOhX-IzGc_8-zjvPTlyYHx_PaXqKM-rkMY95tjCDSJpfVa8xV5T924AKD6EQF5HK8q5MMlS4UsSP3cBuI8vOJ5bzighi0wD2-shb6njcOjMPRIuyn9tiz1t116dAn04Kf6M9UmCQ6xHY4ymWxvwcq-wYXjGQyawcvXxv8wAI833yX1WJKd98Q81RRWoBAzuIIT_23UQrQHZPaB4RlQ5VQrPAaDEAiDXvhEh45WVSHvYqa3vF5MWCOtXFthR1IVJKqdCCTdWCX8PcKCdvX7_1DoGzTSnWDaAJWahprdZ3XpQNs21dYltiGiqsupVOBopFZxxzIS1-46IMVm2fMj44AGoT1sakw1eun2NNKKSIMxvPl4j5HqErGOMpIk2aztYI47KD90YV5hMSMfbqgXzs4PwWdTxseyeUi9yAFvVjcZEi2_y1L78ajtyLlqpXgFeFrbcPVVZexFopzcsqtcwLB0E6JnBWOEsz-4U0fluN_o7m00)
 
-Source: `docs/sources/sequence_open_collection.plantuml`. Created with https://www.planttext.com/
-
+Source: `docs/sources/sequence_open_collection.plantuml`. Created with [planttext](https://www.planttext.com/)
```
david-allison, line 30: Might want to see if we can use GitHub mermaid here

david-allison (commented):
Sounds good! I'm on my phone and will review decisively tomorrow, 

Only question is whether we can move the setup instructions to refer to `docror.sh`

mikehardy:
TIL you should not name your branch the same name as a directory in the top level.
I'm sure there are ways to handle it but easier to recreate.

docs/OVERVIEW.md
mikehardy: yeah - but not right now :-), just doing basic linting right now, not patching much anything

mikehardy:
I actually only saw doctor after!
And then I saw that a lot of what doctor was doing is actually in .github/scripts - so I think there is a lot of DRY violation but it's easy to clean up

This has moved to #173 though

PR #170, opened 2022-04-07
build(deps): integrate gradle updater

mikehardy:
Works well on ankidroid, should work well here? 

May unblock #166

Only difference from Ankidroid is I made this one available on workflow_dispatch as well so I could manually trigger it post-installation to make sure it works


david-allison:
All good from my perspective 👍👍

PR #173, opened 2022-04-08
docs: lint docs, reformat environment doc, specify versions

mikehardy:
This was all required to get things building and to clear the lint out of the MD files (mostly)

supercedes #169 which had a git branch name travesty


mikehardy:
**This should go in after the gradle wrapper update** why? I need to push one change to ENVIRONMENT.md at the end to alter the `-x jar` I put in the example to be the new gradle-7-compatible version

Also, it appears that there is duplication between these files and doctor and the .github/scripts, where this could point to doctor more, and doctor could use the .github/scripts more

david-allison (approved):
Looks great!

PR #178, opened 2022-04-21
build(deps): workflow concurrency limit, JDK11, command-line-tools latest, API31

mikehardy:
This PR began life as an attempt to update sqlite dep from 2.1.0 to 2.2.0, but that needs targetSdkVersion 31

...which needs JDK11
...which needs an up to date set of command line tools
...which alters the way the NDK should install just a bit (pathing, mostly but also simplifies it)

And it required so many workflow runs that I implemented workflow concurrency limits in the process

I'll leave some notes in the actual files where things are interesting, but it's pretty thoroughly tested, will merge if CI is good


.github/scripts/install_ndk.sh
```diff
@@ -1,4 +1,7 @@
 : ${1?"Usage: $0 NDK Version (ex 21.3.6528147)"}
 
 echo "installing NDK $1"
-echo "y" | $(sudo $ANDROID_SDK_ROOT/tools/bin/sdkmanager --install "ndk;$1" --sdk_root=${ANDROID_SDK_ROOT}) | grep -v = || true
+echo $PATH
+sdkmanager --list_installed|grep ndk
```
mikehardy, line 5: - sudo just isn't necessary
- we need a relative PATH to get the new command line tools version of sdkmanager

.github/workflows/linux_build.yml
```diff
@@ -5,6 +5,10 @@ on:
     - cron: "0 0 * * *"
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
mikehardy, line 9: if you implement concurrency limits anywhere else I think this is the simplest definition for it, github.ref is the branch or tag name that triggered it, workflow is the workflow name, and that's the basic unit I want to cancel if a new run is triggered, in all the repos I work on

.github/workflows/update_gradle_wrapper.yml
```diff
@@ -14,11 +18,20 @@ jobs:
...
           submodules: 'recursive'
 
+      - name: Configure JDK 1.11
+        uses: actions/setup-java@v3
+        with:
+          distribution: "adopt"
+          java-version: "11" # minimum for Android API31
+
+      - name: Install Android Command Line Tools
+        uses: mikehardy/setup-android@main
```
mikehardy, line 28: my only reason to fork the action was to update the sdkmanager version it installed, I'll PR that back up stream

build.gradle
```diff
@@ -39,4 +41,20 @@ allprojects {
...
+
+
+ext {
+    jvmVersion = Jvm.current().javaVersion.majorVersion
+    if (jvmVersion != "11" && jvmVersion != "14" && jvmVersion != "16") {
+        println "\n\n\n"
+        println "**************************************************************************************************************"
+        println "\n\n\n"
+        println "ERROR: Anki-Android-Backend builds with JVM version 11, 14, or 16."
+        println "  Incompatible major version detected: '" + jvmVersion + "'"
```
mikehardy, line 54: There was a really subtle error on mac where we ran gradle with shell `bash -l {0}` or something, and that wiped out the PATH so the new JDK wasn't being used and it kept getting the default 8

Would never have known it if it were not for this verification block I implemented originally in AnkiDroid!

mikehardy: :facepalm: just got caught in a different part of the workflow with that bash shell cruft. I grepped and deleted all of it, there is no need in these workflows to specify a shell, it is correct and minimal for us to just run in default shells

Issue #179, opened 2022-04-21
Update NDK / native build tools to latest version


mikehardy:
Right now we're not just using an obsolete / unsupported version (r22) we're using the 22.0 version of it while r22b is out.

But we should move to the latest stable release https://github.com/android/ndk/wiki#current-stable-release

Right now that is r24 / 24.0.8215888

This is useful as part of a general strategy to get everything updated so native M1 builds work as well...
https://github.com/android/ndk/issues/1299#issuecomment-1026338583

That comment lists these things plus some config to do to use correct cmake version

- [x] Android Studio 2021.1.1
- [x] Android Gradle Plugin 7.1.0 (the plugin includes an aapt2 binary)
- [x] Android SDK Build-Tools 32.1-rc1 (SDK's stable channel)
- [x] Android SDK Platform Tools 32.0.0 (SDK's stable channel)
- [x] Android NDK 24.0.8215888
- [x] CMake 3.22.1 (SDK's canary channel)
```

dae:
Is anything remaining here?

mikehardy:
No, I'm going to open a separate issue to bump the compile / target sdk versions but you nailed getting the NDK and toolchain stuff up to date, so M1 native builds should be great now after the #294 epic

PR #180, opened 2022-04-21
build(deps): bump ndk version to r24

mikehardy:
This is the current latest stable version. r23 is LTS, r25 is at beta2

Related #179

This is speculative. It might work. It might not. I'm going to trust CI on it though after the work I've put in on CI just now...

I won't be surprised if this also needs some other chunks of toolchain updated like perhaps build tools, or if it flags new warnings in code and generates some set-warnings-as-errors / fix warnings/errors work. We'll see...

### DIFF
--- a/.github/workflows/linux_build.yml
+++ b/.github/workflows/linux_build.yml
@@ -31,7 +31,7 @@ jobs:
     # COULD_BE_BETTER: Consider turning this into a GitHub action - help the wider community
     # NDK install (unzipping) is really noisy - silence the log spam with grep, while keeping errors
     - name: Install NDK (silent)
-      run: .github/scripts/install_ndk.sh 22.0.7026061
+      run: .github/scripts/install_ndk.sh 24.0.8215888
 
     - name: Install linker
       run: .github/scripts/linux_install_x86_64-unknown-linux-gnu-gcc.sh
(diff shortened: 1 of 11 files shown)


mikehardy:
Total failure :sweat_smile: - will investigate locally as time permits. Not super high priority just wanted to see if it could go in

david-allison:
```
 error occurred: Failed to find tool. Is `/Users/runner/Library/Android/sdk/ndk/24.0.8215888/toolchains/llvm/prebuilt/darwin-x86_64/bin/arm-linux-androideabi-ar` installed?
```

mikehardy:
Oh this is gonna be a rabbit hole - I didn't realize how much infra was contained in the NDK as I'd never really used it before. Hooboy.

So, if you go to r23:

- you will need a workaround for different NDK structure to find the -ar and -clang files (see https://github.com/bbqsrc/cargo-ndk/issues/38 which had a PR but the PR was closed/rejected because it should be fail-fast, should work though in spirt)

- you will need a workaround for missing gcc in the NDK (it was removed after r22) - https://github.com/rust-windowing/android-ndk-rs#supported-ndk-versions - either -Zbuild-std or a libgcc.a that points to libunwind.a or something similar *and* I think you also need a newer Rust (1.59+ ?) *or* Rust's CI moves to a newer Android NDK and gets rid of their last gcc clingon - here is the workaround in code form https://github.com/rust-windowing/android-ndk-rs/pull/189/files#diff-9c2f0d812ce52310e2784d3ba203437e6f318230e4e4f701b321a0f656dfa416R180-R187

No time to play with this yet

Issue #168, opened 2022-04-07
Forward port to Rust 1.62


mikehardy:
Using current rust (currently 1.62) should be possible but we specify 1.58.1 everywhere

to test

```bash
rustup install 1.62.0
rustup default 1.62.0
```

Then make sure you have the cross compilers installed for the android architectures for this version:

```bash
rustup target add armv7-linux-androideabi
rustup target add i686-linux-android
rustup target add aarch64-linux-android
rustup target add x86_64-linux-android
```

should work? just needs testing then updates of doctor + docs + workflows

mikehardy:
cross hasn't released since 2020 (v0.2.1) but has current development, just untagged? Anyway if you use cargo to install cross via git ref, by default it uses tip of default branch which requires cargo with edition 2021 at least

If you pin the cargo install of cross via git to a tag of v0.2.1 it works - also documenting that

entropy!

david-allison:
It's very likely that we'll be able to update the underlying Anki library and get this for free.

I'd rather focus on a fixed target: 2.1.35 works in dev (only one concern with syncing). Once we get that released, we can move to 2.1.49 and it should compile under the latest Rust version

mikehardy:
Perfect. Back-burner on this one then since it does seem to be an rslib problem, we should get it "for free", and what we have now works fine (and is documented! or will be, plus supported by doctor). This can track it just in case though

dae:
This can be closed

mikehardy:
Not until I see the dependency bump actually in and it passes CI everywhere :smiling_imp: - right now doctor and workflows and docs are all saying+using 1.58.1 so there is no demonstrated success

Really, it should say 1.62.0 now anyway as that's current and the idea is to either get current or track why not

dae:
The current code should work fine on the latest Rust, but I used 1.58.1 because that's what the desktop is currently pinned to - there was a compile speed regression at one point that required the pin, though presumably that's been fixed now.

Issue #164, opened 2022-03-16
AnkiDroid unit test suite currently not working on arm64 JDKs on macOS


mikehardy:
**[UPDATE: until solved, we publish `rsdroid-arm64.dylib` on each release in the list of the artifacts and you can point the robolectric tests to it for use on M1 machines]**

----

Symptom:

```
    java.lang.UnsatisfiedLinkError: /private/var/folders/5b/1q9_lqrs79ng7n9hjm_xl4cm0000gn/T/librsdroid2032698931526253765.dylib: dlopen(/private/var/folders/5b/1q9_lqrs79ng7n9hjm_xl4cm0000gn/T/librsdroid2032698931526253765.dylib, 0x0001): tried: '/private/var/folders/5b/1q9_lqrs79ng7n9hjm_xl4cm0000gn/T/librsdroid2032698931526253765.dylib' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64e')), '/usr/lib/librsdroid2032698931526253765.dylib' (no such file)
```

@viciousAegis This is a bug on our side. We do not build `io.github.david-allison-1:anki-android-backend-testing` for M1 Macs.
 
I have an M1 Mac arriving sometime late March, so I'll be personally invested in testing

As a workaround: 
* Remove `@RunWith(AndroidJUnit4::class)`
* Remove `:RobolectricTest()`
* This will only work because the test does not depend on the Anki Database

As a workaround, please enable GitHub actions on your GitHub clone of `Anki-Android` and test there

## Properly fixing this

**Issue**  
* We have a library written in rust: `librsdroid`. This is compiled for Android
* Robolectric doesn't run Android, it runs tests via the JVM on your PC
* Therefore, `librsdroid` needs to be compiled for native platforms
* We have a separate library: `anki-android-backend-testing` which contains the `dlls` and `dylibs` needed
* When this was written, M1 macs didn't exist

Please see: https://github.com/ankidroid/Anki-Android-Backend/tree/main/rsdroid-testing

* First of all, this isn't well documented. Please submit a PR with better documentation if you want
* Most of the work is performed via GitHub actions
* Feel free to submit a pull request to the GitHub actions get this working. 
  * Compile the dylib for `arm64e` (I have no idea how to do this, but it should be possible)
[...]

PR #202, opened 2022-06-16
Update to use Anki 2.1.54's code

### LINKED ISSUE #2: Handle translation loading
The ftl translations aren't currently provided or extracted. These should be readable by the rust

### LINKED ISSUE #17: Re-enable Sync methods
Syncing methods were disabled due to OpenSSL build issues on Windows

We could remove references OpenSSL and a use a rust-based component, or diagnose & patch upstream.

It was likely a path character/space in filename issue.

### LINKED ISSUE #18: @Nullable in backend methods
I noticed in your recent docs that the backend methods are declared with nullable args, eg:

https://github.com/david-allison-1/Anki-Android-Backend/commit/799600f0d487a31137b553cbfb6e90f1a626c79d#diff-427e3c3bcf0309f9e862bb317502e4fc9cb32655a58df4ec9332ed08b36e88c3R79

Was there a particular reason for taking that approach? It is valid protobuf not to set them, and the backend shouldn't fall over if they are not set, but in most cases arguments are expected to be set to something, and a null argument is probably indicative of a bug in the calling code.  The way the Python code does it is to have all arguments of the autogenerated methods be required values, and it then becomes the responsibility of the wrapper code to explicitly provide default values for calls where it makes sense to - perhaps such an approach would work in Java too?

Not a high priority, just wanted to bring it up. And congratulations on getting to this point - it looks like you've put in a mammoth effort.

### LINKED ISSUE #21: chore: Build sqlite.proto in Anki-Android-Backend
This removes an AnkiDroid-specific feature from  `/anki`, which moves it closer to upstream and means we don't need to rebuild the library when making an AnkiDroid specific change.

* Add protobuf generation code into `rslib-bridge/build.rs`
* Enable building Java-based protobufs from Anki-Android-Backend

### LINKED ISSUE #22: Add check to ensure i18n submodule is up to date

### LINKED ISSUE #24: Allow logging to logcat & document usage

### LINKED ISSUE #36: Upgrade to latest Anki backend
Creating issue to keep track of requirements

Indices will start at 0 rather than 1: https://github.com/ankitects/anki/commit/4bd120cc4b9f582034a40a51033548aff3304e56

### LINKED ISSUE #133: Convert to Kotlin
This allows the use of named arguments when converting the Python code.

### LINKED ISSUE #175: [Bug] jacocoTestReport orphans files on windows
###### Reproduction Steps

1. run `.\gradlew.bat jacocoTestReport` on windows with an emulator up (may only require unit tests, but this runs all tests)
2. observe `%APPDATA%\local\temp`
3. 

###### Expected Result

No files left over after test runs, successful tests or not

###### Actual Result

Lots of rsdroidNNNNNNN.dll files, other random test files scattered about
They consume a few GB per test run so they will fill up your disk quickly if you run a lot of unit tests...

###### Debug info

tip of main branch as of this typing

###### Research
*Enter an [x] character to confirm the points below:*

- [x] I have read the [support page](https://ankidroid.org/docs/help.html) and am reporting a bug or enhancement request specific to AnkiDroid
- [x] I have checked the [manual](https://ankidroid.org/docs/manual.html) and the [FAQ](https://github.com/ankidroid/Anki-Android/wiki/FAQ) and could not find a solution to my issue
- [x] I have searched for similar existing issues here and on the user forum
- [x] (Optional) I have confirmed the issue is not resolved in the latest alpha release ([instructions](https://docs.ankidroid.org/manual.html#betaTesting))

### LINKED ISSUE #209: Update OVERVIEW / diagrams to match new backend state
> Before I spend much more time on this, I'd appreciate it if you could let me know whether you're amenable to the changes in [b02515b](https://github.com/ankidroid/Anki-Android/commit/b02515b0f6a403a4e9f278d21620a1b07d0b1232). I was planning to build upon them to look into adding some new backend functionality, and don't really relish the thought of having to add all the extra boilerplate to do so.

I am personally okay with it, in the first instance by criteria of "implementor's choice" but also by scanning the overview doc and reasoning about the initial motivation:

https://github.com/ankidroid/Anki-Android-Backend/blob/main/docs/OVERVIEW.md#droidbackendrustdroidbackend

> Allows a testable comparison between the Java and the Rust during the conversion

I think this was useful but may have served it's purpose. We're definitely going to use rslib. The babel's tower of build toolchains and languages in use is a bit baroque, but we're using it live, we're not going back. How much value does a comparison with a java implementation that's never going to keep up with rslib development have? My opinion, none in future

> Allows a pure Java conversion afterwards (if deemed appropriate due to AGPL concerns)

I don't think we have those concerns? I do not have those concerns

> Encapsulates all access to the Rust, allowing the implementation to later be swapped out within one file.

Similar to first statement above: swapped out with what? I don't intend to get into an Anki development competition with rslib development so there will never be a full-featured swapout

https://github.com/ankidroid/Anki-Android-Backend/blob/main/docs/OVERVIEW.md#rustbackend-interface

One point here is interesting:

> Allows for easy mocking, and method discovery

Is easy mocking for testability interesting? Is this even the appropriate layer for testing the code - that is who would be mocking out RustBackend, is it a facility actually used now, and would any change proposed here alter that? I don't have a feel for it

In general, if this goes ahead it would be good to update the docs / diagrams for future intrepid explorers: 

> Source: docs/sources/sequence_method_call.plantuml. Created with [planttext](https://www.planttext.com/)

I personally don't see a problem. We're going to use rslib, no way around it. We either commit (which again: is not user hostile, I just scanned exception reports again) or are in a dev race with rslib which is a non-starter. The interfaces are code-gen'd so just calling in as simply as possible seems right to me.

_Originally posted by @mikehardy in https://github.com/ankidroid/Anki-Android/issues/11644#issuecomment-1159485319_

dae:
Requires the following branch: https://github.com/ankidroid/anki/pull/2

- Updates David's previous work to work with the latest desktop code.
- Most of the Rust code has been moved into the branch above; only about
  120 lines of Rust remain here, to implement the small JNI interface. The
  AnkiDroid-specific routines flow through the same backend call interface
  as the other backend code, with errors handled correctly.
- Migrates the code to Kotlin. The tests remain in Java - I couldn't
  figure out how to get the Kotlin plugin working on the tests. The conversion
  introduced some warnings on compile that should probably be addressed at one
  point.
- Translations are now included as part of the build, including statically-typed
  accessors in Kotlin.
- The code generation from the proto files now targets Kotlin, and marks
  input arguments as non-optional unless they've been explicitly declared
  as nullable in the proto files.

Some design notes:

- Recent Anki versions use `anki.<module>` as the package path for
  generated protobuf, and I've kept this for the Kotlin code as well (
  eg `import anki.notes.Note` to import the `Note` message defined in proto/anki/notes.proto).
  The rsdroid module is still using net.ankiweb.rsdroid; not sure if you want
  that changed or kept the same.
- Generated methods use camelCase: https://github.com/ankidroid/Anki-Android/issues/11582
- I've incremented the version number. One possible alternative to semver that might be
  worth considering in the future is to encode the desktop version number in the release
  instead, eg 2.1.53 -> 21.53.0 (last number for patch/revision).
- https://github.com/ankidroid/Anki-Android-Backend/blob/f9df8855de8fd7e322a39ec7b38356a80381b6a0/rsdroid/src/main/java/net/ankiweb/rsdroid/BackendMutex.java has been dropped. Almost all backend calls flow through a single path, and
  the locking can be handled there instead. Only the transaction methods handle the locking
  specially.
- Like the desktop code, instantiating a Backend object now instantiates the underlying
  Rust object at the same time, instead of requiring a two stage initialization.

There are some warnings showing that were introduced as part of the Kotlin conversion,
that can be fixed in a future commit.

Closes #209
Closes #175
Related #168 (was 'closes' but it currently stops on 1.58.1, so does not close although moves us from rust 1.54) 
Closes #133
Closes #36
Closes #24
Closes #22
Closes #21 
Closes #18
Closes #17
Closes #2


README.md
```diff
@@ -20,8 +20,8 @@ Adapter allowing AnkiDroid to leverage Anki Desktop's Rust-based business logic
 Pre-built version:
 
 ```gradle
-    implementation "io.github.david-allison-1:anki-android-backend:0.1.11"
-    testImplementation "io.github.david-allison-1:anki-android-backend-testing:0.1.11"
+    implementation "io.github.david-allison-1:anki-android-backend:0.1.12"
+    testImplementation "io.github.david-allison-1:anki-android-backend-testing:0.1.12"
```
mikehardy, line 24: Hmm - thinking just about your versioning brainstorm, and the current version

> I've incremented the version number. One possible alternative to semver that might be
worth considering in the future is to encode the desktop version number in the release
instead, eg 2.1.53 -> 21.53.0 (last number for patch/revision).

1- I dislike "magic zero" and "<1" version numbers, just smells like someone lacks confidence and is wasting a semantically-useful slot, so I think we should be >=1 here regardless
2- I like the idea of our backend version containing information about the upstream version since that's *really* what we're built on and most of what the resulting artifact actually is, so I like the idea of using the upstream version as the base for anything, but I don't want it to be some "oh if you know we convert it it makes sense you just combine the ..." shenanigans - I like things that are stupid-obvious if possible - meaning we should just start with 2.1.53 here I think
3- I think it's important to preserve our own versioning capability, but no problem, I think semantic-versioning's pre-release capability allows for that. So we should be "2.1.53-1" here


11.4 here https://semver.org/#spec-item-11 - and it appears to contain enough semantic power to effectively encode an entire sub-semantic release series after the hyphen if needed though a simple digit for now should be fine based on our past versioning here

mikehardy:
@david-allison 

> Could we limit the translations to the following, as that should produce a well-needed reduction in apk size and serves as a stopgap before we think about combining CrowdIn and Pontoon


**should produce** --> premature optimization. what's the size of the translations + size per translation + expected reduction?

README.md
mikehardy: :point_up: works for me

dae:
Translation size depends on how many strings are translated - almost full/almost empty examples:

```
[dae] code/droid/Anki-Android-Backend [new-anki]  % dua -A ftl/core/core/fr
  91.35 KB ftl/core/core/fr
[dae] code/droid/Anki-Android-Backend [new-anki]  % dua -A ftl/core/core/en-GB
   1.13 KB ftl/core/core/en-GB
```

AnkiMobile has fewer active languages, and I don't bother to strip them there. The key names are shared across languages, and likely compress well.

> (For discussion) I'd prefer to stick to the 'standard' Java convention for package names in the generated code. I don't care what that is, if we want to move away from ankiweb

My thinking here is was that if you see a class imported from 'anki.*.*', you can tell it's a DTO from the backend, and easily look it up in the .proto files or desktop code. And `anki.decks.Deck` is somewhat less awkward than `net.ankiweb.anki.decks.Deck`. 
There are also other existing examples that don't follow reverse DNS notation, eg `import java.util.Arrays`, `import androidx.sqlite.db.SupportSQLiteOpenHelper` and so on.

dae:
The dependabot changes caused a conflict; I've rebased over them and updated the remaining deps in the last commit.

mikehardy:
Readme and easy testing docs for version bump too? Hoping to work through local verification and integration if these this weekend, are you happy with them or should I wait? Just lemmeno

dae:
If you guys are happy with them, they should be ok to merge - with a new release out, the draft PR on the AnkiDroid end can run through CI (passes the unit tests locally here)

david-allison (approved):
I'm a little uneasy with `anki` as the package name, but the arguments make sense, and it's senseless to bikeshed. Happy my other comments have been handled.

Let's do this. Thanks so much!

dae:
If you guys change your mind about the naming later, that should be fairly easy to adjust with a find&replace.

I've gone through the issue backlog and linked the ones that this PR should close.

mikehardy:
Oh, that windows temp dir commit is gold - seriously thank you!

dae:
Fixed instructions in easy-testing; the value needs to be set in the app onCreate, not deck picker. Might be worth holding off on this for a few days, as I'll have a chance to give it some testing as I experiment with adding some new functionality.

rsdroid/src/main/java/net/ankiweb/rsdroid/Backend.kt
```diff
@@ -174,7 +174,7 @@ open class Backend(val context: Context, langs: Iterable<String> = listOf("en"),
     @CheckResult
     override fun fullQuery(query: String, bindArgs: Array<Any?>?): JSONArray {
         return try {
-            Timber.i("Rust: SQL query: '%s'", query)
+            Timber.d("Rust: SQL query: '%s'", query)
```
mikehardy, line 192: probably a noticeable performance boost by tamping this down, may be one of the few places where it makes sense to to check log level before doing the thing even, to avoid memory allocation. I think Timber will avoid the string formatting if a level isn't in use but the allocs themselves may be substantial. I looked and I'm not sure Timber's loggable check is implemented in the tree we plant - I think we would need to subclass it and otherwise it just defers to android logger, but a BuildConfig.DEBUG conditional could work. Could be premature optimization

mikehardy: this is where we override in ankidroid FWIW - I don't think it's worth it here vs a pure cutout IMHO https://github.com/ankidroid/Anki-Android/blob/5f2ca812067c0a0e249e4dffce488c184d201ccd/AnkiDroid/src/main/java/com/ichi2/anki/AnkiDroidApp.java#L426

dae:
With these latest changes, basic sync support is now available to play with on https://github.com/ankidroid/Anki-Android/pull/11644

rsdroid/src/main/java/net/ankiweb/rsdroid/Backend.kt
dae: It's probably not great for performance, but it was also noisy :-) IMHO this logging might be better done on AnkiDroid's side, and perhaps only when required.

mikehardy:
> Might be worth holding off on this for a few days, as I'll have a chance to give it some testing as I experiment with adding some new functionality.

Do you still feel like this is the case? Today is probably the best chance for me to test integrate this stuff locally and merge it, for the next few days or week. And I've got a strong aversion to letting large cross-cut style changes sit since there is so much work in flight here right now. We accumulate conflicts pretty quickly otherwise...

dae:
I wrote that when I started working on syncing and found it crashing immediately :-) This side seems to be working pretty well, and I can't see much if any downsides to merging it. The associated changes in the AnkiDroid repo are a bit riskier - while I have tried to minimize the changes to the code paths that are active by default and unit tests are passing locally, I haven't spent much time actually using the app with these changes, so you might want to at least give it a brief test on device before pulling the trigger.

mikehardy:
Perfect

> so you might want to at least give it a brief test on device before pulling the trigger.

Yep, that was going to be the majority of the effort I think. Using it against a collection I'm familiar with in legacy + V16+ modes on an emulator then assuming it passes doing same on a real device. I'm excited to see it

mikehardy:
Just to perhaps reset expectations just a touch - took me a bit longer to get a major react-native-firebase than I wanted so this may fade a couple days but my excitement to get the entire thing launched hasn't faded. So I don't think I was sensing impatience except on my own behalf but if it existed, temper it a just a bit, but we'll get this and all related work on the anki fork and AnkiDroid out

dae:
I know how these things go! No big rush on my end.

mikehardy:
:boom: https://github.com/ankidroid/anki/pull/2 is merged, next up is this review+merge+release of this one then I'll move up another layer to AnkiDroid

gradle.properties
```diff
@@ -19,7 +19,7 @@ android.useAndroidX=true
 android.enableJetifier=false
 
 GROUP=io.github.david-allison-1
-VERSION_NAME=0.1.11
+VERSION_NAME=0.1.13-anki2.1.53
```
mikehardy, line 22: Just curious if there is a reason it went to 0.1.13-nnn vs 0.1.12-nnn ? speculative suggestion based on the thought

```suggestion
VERSION_NAME=0.1.12-anki2.1.53
```

mikehardy:
> boom [ankidroid/anki#2](https://github.com/ankidroid/anki/pull/2) is merged, next up is this review+merge+release of this one then I'll move up another layer to AnkiDroid

I think this implies you need to update the commit hash for the git submodule in rslib? Or I can - I'll be AFK for a few hours and it's no big deal either way. Cheer Damien

dae:
I think the submodule probably is fine as is - when I opened the PR on your repo it adds that sha1 into the repo at PR creation time, so I was able to point git to it even though you hadn't merged it.

gradle.properties
dae: I bumped it once at the start, and again after I changed the backend API, just to make it clearer that the API had changed. Feel free to adjust as desired :-)

mikehardy: okay, doesn't really matter in the grand scheme, version numbers are free and every commit has a cost, so we go with 13, done!

mikehardy:
> I think the submodule probably is fine as is - when I opened the PR on your repo it adds that sha1 into the repo at PR creation time, so I was able to point git to it even though you hadn't merged it.

:thinking: I can see how technically that's 100% correct - I'm thinking for transparency and build reproducibility blah blah I'll update the hash so it's pointing to the official one now 

A bit later from the above sentence as I was working on it: stupid git question, I'm having a heck of a time checking out this branch correctly with submodules (specifically: ftl/core) intact, either with `gh` command line tool or with ankitects added as a remote and `git checkout new-anki --recurse-submodules` - I'm sure I'll find the answer on the internet but perhaps you know off the top of your head the magic incantation to go from `main` here to this branch on the command line ?

dae:
The action workflows use 'git submodule update --init --recursive' - does that not work?

mikehardy:
> The action workflows use 'git submodule update --init --recursive' - does that not work?

Nope, saw that (as well as the github actions transcripts) and tried it along with quite a few variations a little poking around in $gitroot/.git/modules etc

```
mike@bistromath:~/work/AnkiDroid/Anki-Android-Backend ((5d29116...) *) % git submodule update --init --recursive
fatal: not a git repository: /home/mike/work/AnkiDroid/Anki-Android-Backend/ftl/core/../../.git/modules/ftl/core
Failed to clone 'ftl/core'. Retry scheduled
BUG: submodule considered for cloning, doesn't need cloning any more?
fatal: could not get a repository handle for submodule 'ftl/core'
fatal: not a git repository: ftl/core/../../.git/modules/ftl/core
mike@bistromath:~/work/AnkiDroid/Anki-Android-Backend ((5d29116...) *) % 
```

I'll just head over to stackoverflow like everyone else (or gitbook) and learn something new, no worries

dae:
Maybe you have old files lying around?

```
[dae] Work/code/droid  % mkdir tmp
[dae] Work/code/droid  % cd tmp
[dae] code/droid/tmp  % git clone https://github.com/ankitects/Anki-Android-Backend.git -b new-anki
Cloning into 'Anki-Android-Backend'...
remote: Enumerating objects: 3348, done.
remote: Counting objects: 100% (1409/1409), done.
remote: Compressing objects: 100% (542/542), done.
remote: Total 3348 (delta 958), reused 1078 (delta 759), pack-reused 1939
Receiving objects: 100% (3348/3348), 788.22 KiB | 5.75 MiB/s, done.
Resolving deltas: 100% (1793/1793), done.
[dae] code/droid/tmp  % cd Anki-Android-Backend
[dae] droid/tmp/Anki-Android-Backend [new-anki]  % git submodule update --init --recursive
Submodule 'ftl/core' (https://github.com/ankitects/anki-core-i18n.git) registered for path 'ftl/core'
Submodule 'anki' (https://github.com/ankidroid/anki) registered for path 'rslib-bridge/anki'
Cloning into '/home/dae/Work/code/droid/tmp/Anki-Android-Backend/ftl/core'...
Cloning into '/home/dae/Work/code/droid/tmp/Anki-Android-Backend/rslib-bridge/anki'...
Submodule path 'ftl/core': checked out '460c0d2954e40513e6cb65b206fed6412cee5817'
remote: Enumerating objects: 8, done.
remote: Counting objects: 100% (2/2), done.
remote: Compressing objects: 100% (2/2), done.
remote: Total 8 (delta 0), reused 2 (delta 0), pack-reused 6
Unpacking objects: 100% (8/8), 2.06 KiB | 2.06 MiB/s, done.
From https://github.com/ankidroid/anki
 * branch                66d3cbe41018766411c38577d56671922d798d56 -> FETCH_HEAD
Submodule path 'rslib-bridge/anki': checked out '66d3cbe41018766411c38577d56671922d798d56'
```

dae:
> I can see how technically that's 100% correct - I'm thinking for transparency and build reproducibility blah blah I'll update the hash so it's pointing to the official one now

I was assuming you'd do a simple git pull to get the commits. The way you merged the changes seems to have added 'committed by Mike Hardy' to them, which changed the hash; if it had stayed the same, the sha1 should have been identical. Agreed that it makes sense to update it given that the hash changed.

mikehardy:
> I was assuming you'd do a simple git pull to get the commits. The way you merged the changes seems to have added 'committed by Mike Hardy' to them, which changed the hash; if it had stayed the same, the sha1 should have been identical. Agreed that it makes sense to update it given that the hash changed.

Yeah - rebase merge does that, but also has that nice linear commit history. Tradeoffs, and I blame Tim Rae for the rebase merge habit ;-). Anyway something's off with local state and `git clean -d -x -f` didn't blow it out but I'll figure out, I'm not letting some silly submodule thing block me for long

rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/database/DatabaseRegularCorruptionTest.java
```diff
@@ -35,8 +36,12 @@ public class DatabaseRegularCorruptionTest extends DatabaseCorruption {
     protected void assertCorruption(Exception setupException) {
         // Rust: net.ankiweb.rsdroid.BackendException$BackendDbException: DBError { info: "SqliteFailure(Error { code: DatabaseCorrupt, extended_code: 11 }, Some(\"database disk image is malformed\"))", kind: Other }
         // Java: database disk image is malformed (code 11): , while compiling: PRAGMA journal_mode
+
+//        assertThat(setupException.getClass(), typeCompatibleWith(BackendException.BackendDbException.class));
         assertThat(setupException.getClass(), typeCompatibleWith(SQLiteDatabaseCorruptException.class));
 
+        // this mapping to an unrelated exception should be done at a higher level
```
mikehardy, line 43: this mapping to an unrelated exception should be done at a higher level

Do you think you could use the github context menu on this comment to peel this to a new issue and specify exactly what the design issue is here?

rsdroid/build.gradle
```diff
@@ -72,7 +80,25 @@ dependencies {
...
+}
+
+preBuild.dependsOn "generateTranslations"
+
+Boolean wantAllPlatforms = System.getenv("CURRENT_ONLY") != "true"
+
+if (wantAllPlatforms) {
+    preBuild.dependsOn "cargoBuild"
+} else {
+    preBuild.dependsOn "cargoBuildX86_64"
```
mikehardy, line 100: shouldn't this determine the correct platform? Seems like cargoBuildX86_64 is an assumption as what the opposite of wantAllPlatforms is. David has an M1 now, for instance. That may be fancy so perhaps a middle-ground would be to failfast if CURRENT_ONLY == true and arch != x86-64 (or linux/x86-64 if it's that specific)

rsdroid/proto.gradle
```diff
@@ -35,6 +33,7 @@ protobuf {
     }
     generateProtoTasks {
         all().each { task ->
+            task.outputs.upToDateWhen { false }
```
mikehardy, line 36: do I read this correctly as "run these tasks every time no matter what?" Is there a reason freshness / cache-ability cannot be determined for them?

rsdroid/src/main/java/net/ankiweb/rsdroid/BackendFactory.kt
```diff
@@ -0,0 +1,74 @@
...
+    fun setDefaultLanguagesFromLocales(locales: Iterable<Locale>) {
+        defaultLanguages = locales.map { localeToBackendCode(it) }
+    }
+
+    private fun localeToBackendCode(locale: Locale): String {
+        // TODO: this needs checking that all language codes match the ones
+        // shown here: https://i18n.ankiweb.net/teams/
+        return when (locale.language) {
+            Locale("heb").language -> "he"
+            Locale("yue").language -> "zh-TW"
```
mikehardy, line 65: We have 4 of these fancy ones, do you have the other two?

https://github.com/ankidroid/Anki-Android/blob/6ee8892b1309e3116d52d512696b5fbb01c729ba/tools/update-localizations.py#L201-L208

```python
    if language == 'yu':
        androidLanguage = 'yue'
    if language == 'he':
        androidLanguage = 'heb'
    if language == 'id':
        androidLanguage = 'ind'
    if language == 'tl':
        androidLanguage = 'tgl'
```

tools/get-buildinfo.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+#
+# The buildinfo.txt file should be generated as part of the build,
+# but for now we'll just check it into source control.
+
+(cd rslib-bridge/anki && bazel build -c opt buildinfo.txt)
+cp rslib-bridge/anki/.bazel/bin/buildinfo.txt rslib-bridge
```
mikehardy, line 7: For a bazel noob like myself might be good to do the equivalent of a 

```bash
if ! bazel --version /dev/null 2>&1; then
  echo "Hey noob, you need to run the bazel build to do this so go apt yourself some bazelisk (correct package name? ) and run this that the other"
  exit 1
fi
```

.gitmodules
```diff
@@ -2,3 +2,6 @@
 	path = rslib-bridge/anki
 	url = https://github.com/ankidroid/anki
 	branch = ankidroid-2-1-34
```
mikehardy, line 4: this somehow seems wrong, however the commit hash should be (is) authoritative. Still, why is this branch in there and why would it not be ankidroid-2.1.53 now :thinking:

docs/TESTING.md
```diff
@@ -0,0 +1,79 @@
...
+- rustup target add x86_64-linux-android
+- sudo ln -sf /usr/bin/gcc /usr/bin/x86_64-unknown-linux-gnu-gcc
+
+Install protobuf:
+
+- Install protobuf with your package manager
+
+## Optional Python venv
+
+If you don't want to `pip install protobuf` globally, you can
```
mikehardy, line 32:
```suggestion
If you don't want to `pip install` the stringcase and protobuf python packages globally, you can
```

rsdroid/build.gradle
dae: I was hoping @david-allison would swoop in and add code to handle that case, as I don't have AS set up on my Mac to test with.

docs/TESTING.md
```diff
@@ -0,0 +1,79 @@
...
+symlink the python bin from a venv into `python` at the top of
+the project folder:
+
+```
+$ ln -sf /path/to/venv/bin/python python
```
mikehardy, line 37: do  you need to symlink the pip into there as well? or is pip smart enough to find the working dir of the venv python? I've done virtualenv python stuff before but only would say I know it in passing

rsdroid/proto.gradle
dae: I seem to recall that changes to the .proto files were not triggering a rebuild. Presumably it could be made to check the input files, but I don't know how to do that, and the build time is not particularly long at the moment. :-)

rsdroid/src/main/java/net/ankiweb/rsdroid/BackendFactory.kt
dae: No translations for those two languages yet, but I've added them to the code.

mikehardy (commented):
I peppered comments here and there since I thought I might get you in AUS (NZ?) morning in case I didn't finish the review. Finished reading it all now, nothing big at all IMHO

I've got it building locally and I'll do some test integrations but as ever, leaning towards merge as fast as I can process change, though I am curious about your thoughts on the comments

.gitmodules
dae: You can probably remove it, I think it's only applied when using --remote with git submodule.

docs/TESTING.md
dae: Nah,  you need to run the pip command from the venv - I've tweaked the text to clarify.

rsdroid/build.gradle
mikehardy: I have an M1 I can check on, I'll peel it to an issue and in the meantime I'll think of it as "the optional feature you made doesn't work in all cases" - in other words, not a blocker so long as it's tracked

mikehardy:
Since you have this up and running right now and I would hate to mess with someone else's branch that's being worked, do you think you could try the rslib/anki commit hash re-point and check the .git/modules branch thing? Those are the only two nits I can think of to pick. Also, if they don't work we can bail then from a review perspective this is all good

Those won't affect the build either, so I can focus on test integrations here with the local.properties changes on AnkiDroid side to use local build and toggle legacy on and off

dae:
Feel free to push changes into this branch if you need to, it's easy enough to rebase over.

david-allison:
One resolved error (to be peeled off into a docs update). 

One error which is pending
Note:`. "$HOME/.cargo/env"` is in my `.profile` and `.zprofile` and `cargo` runs in a shell

```
Compilation is not supported for following modules: Anki-Android-Backend. Unfortunately you can't have non-Gradle Java modules and Android-Gradle modules in one project
```

```
Traceback (most recent call last):
  File "./tools/genfluent/genfluent.py", line 31, in <module>
    modules = get_strings()
  File "./tools/genfluent/genfluent.py", line 26, in get_strings
    subprocess.run(["cargo", "run", output_file], check=True, cwd="rslib-bridge/anki/rslib/i18n")
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/subprocess.py", line 493, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/subprocess.py", line 858, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/subprocess.py", line 1704, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'rslib-bridge/anki/rslib/i18n'
```

Needed a submodule update

```
Traceback (most recent call last):
  File "./tools/genfluent/genfluent.py", line 31, in <module>
    modules = get_strings()
  File "./tools/genfluent/genfluent.py", line 26, in get_strings
    subprocess.run(["cargo", "run", output_file], check=True, cwd="rslib-bridge/anki/rslib/i18n")
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/subprocess.py", line 493, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/subprocess.py", line 858, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/subprocess.py", line 1704, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'cargo'
```

dae:
genfluent.sh uses /bin/bash; is .cargo/env on the PATH of the parent process? If you're loading AS from the macOS GUI, it may not have inherited path changes if they only apply to login shells

mikehardy:
Hmm - I opened in Android Studio on Ubuntu with no issue. Haven't tried in my x86-64 osx-kvms yet or on my wife's m1 machine which I have to ask for time slices on  ;-). Maybe mac specific but that is unexpected. I'm out of time for the day though - this is my last out-the-door comment then I'll try again my Monday morning

dae:
Note you'll need to do a ./gradlew clean after the latest commits, as the protobuf generation doesn't remove old files when generating new ones, and it appears to have broken the build in this case.

david-allison (approved):
(my approval still stands)

mikehardy:
Verified the rslib/anki commit hash is 2f7a0202976b3a400542b0a9bf5312c0e633472a and that hash is both the actual hash on ankidroid/anki#ankidroid-2.1.54 and the one committed here. So the bump to that looks clean :rocket: - back to checking integration here then merge + release

Something is rough on the whole submodule thing though. I basically just deleted my clone and made a new one in order for all the submodules to work well. Not a huge deal for me and probably a local problem but just noting it in case others get stuck. Killing a local clone and making a new one isn't that painful and it's a O(1) operation - always just a few minutes - vs troubleshooting

mikehardy (approved):
my last item was a "nice to have" on rslib/anki commit hash and the submodule thing. Both conclusively nailed. Let's launch this sucker

mikehardy:
It's a lot of commits but I like the commit history here, nearly all of them are important, I believe they all actually worked independently (so bisect will work) but we don't bisect a lot here and this change is so huge that bisect wouldn't be too useful before/after anyway. So I'm rebase merging

mikehardy:
1- tagged it 0.1.14-anki2.1.54 to match version in properties file, pushed the tag
2- made a release from the tag with auto-generated release notes
3- built Apple Silicon dylib following #164 and added it to the release
4- kicked off the two sonatype publish builds and I'll log in to finalize them in half an hour or so after they are built

PR #217, opened 2022-06-29
Build anki web assets from rslib-bridge/anki submodule

krmanik:
There is need to generate web assets file for showing graphs, card info, import csv etc. The file built using `bazel`, so it needs to integrate in backend code gradle build.

The current PR will solve the issue of generating assets for current anki version the this (backend) repo

The flow of this PR
1. Use actions to run `build-web-assets.sh`
2. First pull `submodule` of latest anki, (can be changed to pull specific commit hash from buildinfo.txt but the actions will run when `buildinfo.txt` changed.)
3. The web assets built using `bazel` then copied to `anki-web-assets` branch in a folder `web`.
4. In next PR, the following raw url will be used to build the backend with `resources` in build.gradle
```
https://raw.githubusercontent.com/krmanik/Anki-Android-Backend/anki-web-assets/web/anki_artifacts.zip
```
5. Next PR will solve the issues
Issues need to fixes
```
This has a few issues that need resolving before it's usable:

- Need to tell Gradle that unpackAnkiArtifacts should be run prior to the cargo lines
- downloadAndPrepareArtifacts chokes on clean build because rsdroid/build does not exist; probably needs creating if missing
```


dae:
While committing into a separate branch of the same repo is a common pattern for actions like doc generators, the artifacts will be multi megabytes, so it will cause the size of this repo to grow in size - I suggest the action publishes an artifact instead.

krmanik:
I will update it publish the artifacts.

krmanik:
After changing some `ts` files in `anki` submodule for testing, the local built `aar` file should contains latest `ts` file. The current implementation in this PR will not update the new web assets in `aar` file. So, I came up with following implementation, if environment already configured to run `bazel` command, then it can be used to build the web assets. First run will take some time, but next run of bazel command will be faster. Also clean command works without any issues. 

```diff
diff --git a/rsdroid/build.gradle b/rsdroid/build.gradle
index fd7bb40..12f8736 100644
--- a/rsdroid/build.gradle
+++ b/rsdroid/build.gradle
@@ -45,10 +45,23 @@ android {
     sourceSets {
         main {
             kotlin.srcDirs += "build/generated/source/fluent"
+            resources {
+                srcDirs "src/main/resources", "build/generated/anki_artifacts"
+            }
         }
     }
 }

+task buildAnkiWebAssets(type: Exec) {
+    workingDir "$rootDir"
+    String toolPath = System.getProperty('os.name').toLowerCase(Locale.ROOT).contains('windows') ? 'tools\\web\\build-web-assets.bat' : 'tools/web/build-web-assets.sh'
+    if (System.getProperty('os.name').toLowerCase(Locale.ROOT).contains('windows')) {
+        commandLine 'cmd', '/c', toolPath
+    } else {
+        commandLine 'sh', '-c', toolPath
+    }
+}
+
 // Consider upgrade to DSL: https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block
 apply plugin: "org.mozilla.rust-android-gradle.rust-android"

@@ -96,6 +109,8 @@ task generateTranslations(type: Exec) {
     }
 }

+preBuild.dependsOn "buildAnkiWebAssets"
+
 preBuild.dependsOn "generateTranslations"

 Boolean wantAllPlatforms = System.getenv("CURRENT_ONLY") != "true"
diff --git a/tools/web/build-web-assets.sh b/tools/web/build-web-assets.sh
new file mode 100755
index 0000000..d2329d4
--- /dev/null
+++ b/tools/web/build-web-assets.sh
@@ -0,0 +1,17 @@
+# build web assets with anki submodule in rslib-bridge
+
+SRC_DIR=$(pwd)
+ANKI_SRC=$SRC_DIR/rslib-bridge/anki
+BUILD_DIR=$SRC_DIR/rsdroid/build/generated/anki_artifacts/
+
+# build for ts
+cd $ANKI_SRC/ts
+bazel build ts
+
+# build for web
+cd $ANKI_SRC/qt/aqt/data/web
+bazel build web
+
+# copy pages to build/generated
+mkdir -p $BUILD_DIR/web
+cp -r $ANKI_SRC/.bazel/bin/qt/aqt/data/web/pages/* $BUILD_DIR/web
```

dae:
This is convenient for testing, and is probably the best long term approach, but the problem in the short term is that it will require work to make sure the Bazel build works in all the different github actions (eg Windows/Mac build)

dae:
One other issue is that the .bazel folder will make AndroidStudio really slow, so you may want to manually specify a different location, or remove the symlink as part of the build.

krmanik:
> This is convenient for testing, and is probably the best long term approach, but the problem in the short term is that it will require work to make sure the Bazel build works in all the different github actions (eg Windows/Mac build)

I will update the github actions workflows file for other system.

> One other issue is that the .bazel folder will make AndroidStudio really slow, so you may want to manually specify a different location, or remove the symlink as part of the build.

I have used `--symlink_prefix` option in bazel command to symlink .bazel folder in tmp dir. The only issues is `node_modules`, the current size of `node_modules` is 450MB approx. I have tried to find the solution for getting `node_modules` from /tmp folder but didn't found any. May be the anki source should be copied to temp dir. 

This is working solution without considering `node_modules` size.
```sh
# build for ts
cd $ANKI_SRC/ts
bazel build ts --symlink_prefix=/tmp/.bazel/

# build for web
cd $ANKI_SRC/qt/aqt/data/web/pages
bazel build pages --symlink_prefix=/tmp/.bazel/

# copy pages to build/generated
mkdir -p $BUILD_DIR/web
cp -r /tmp/.bazel/bin/qt/aqt/data/web/pages/* $BUILD_DIR/web
```

I have also tried to symlink the anki dir (`sudo ln -s $ANKI_SRC $TMP_DIR`), but .bazel is still in the main dir not in temp dir.

dae:
That might be enough already? If node_modules is causing a problem, perhaps there's a way to exclude it in AndroidStudio that can be configured in a file in the repo?

krmanik:
Will it be considered in build.gradle?
```
apply plugin: 'idea'

idea.module {
    excludeDirs += file('node_modules/')
}
```
https://stackoverflow.com/questions/47244841/how-do-i-exclude-a-directory-like-node-modules-from-android-studios-open-fil

krmanik:
I will reopen it because I have renamed this branch so that when I push the github action should run. But it get closed.

tools/web/build-web-assets.bat
```diff
@@ -0,0 +1,17 @@
+:: build web assets with anki submodule in rslib-bridge
+
+set SRC_DIR=%cd%
+set ANKI_SRC=%SRC_DIR%\rslib-bridge\anki
+set BUILD_DIR=%SRC_DIR%\rsdroid\build\generated\anki_artifacts
+
+:: build for ts
+cd %ANKI_SRC%\ts
+bazel build ts --symlink_prefix=%temp%\.bazel\
```
dae, line 9:
```suggestion
bazel build ts/... --symlink_prefix=%temp%\.bazel\
```

Otherwise it's not actually building everything. Same issue in .sh file

dae: Actually, that step may be unnecessary if you're building the pages below, since they should automatically cause the ts components to be built.

There's only one part from ts that will be needed:

bazel build ts/reviewer/reviewer_extras_bundle
cp .bazel/bin/ts/reviewer/reviewer_extras_bundle.js 

AnkiDroid will want to include that in the reviewer js in the future in order to support custom scheduling in the v3 scheduler.

tools/web/build-web-assets.sh
```diff
@@ -0,0 +1,17 @@
...
+cd $ANKI_SRC/ts
+bazel build ts --symlink_prefix=/tmp/.bazel/
+
+# build for web
+cd $ANKI_SRC/qt/aqt/data/web/pages
+bazel build pages --symlink_prefix=/tmp/.bazel/
+
+# copy pages to build/generated
+mkdir -p $BUILD_DIR/web
+sudo cp -r /tmp/.bazel/bin/qt/aqt/data/web/pages/* $BUILD_DIR/web
```
dae, line 17:
```suggestion
cp -r /tmp/.bazel/bin/qt/aqt/data/web/pages/* $BUILD_DIR/web
chmod -R a+w $BUILD_DIR/web
```
rm the existing folder to fix the problem, then it should work going forward

dae:
One remaining nice-to-have would be to generate the buildinfo.txt file, so it doesn't need to be checked into this repo/manually updated.

tools/web/build-web-assets.sh
```diff
@@ -1,17 +1,33 @@
...
 ANKI_SRC=$SRC_DIR/rslib-bridge/anki
 BUILD_DIR=$SRC_DIR/rsdroid/build/generated/anki_artifacts
 
-# build for ts
-cd $ANKI_SRC/ts
-bazel build ts --symlink_prefix=/tmp/.bazel/
+# generate buildinfo.txt file
+cd $ANKI_SRC
+bazel build --symlink_prefix=/tmp/.bazel/ -c opt buildinfo.txt
+cp /tmp/.bazel/bin/buildinfo.txt $SRC_DIR/rslib-bridge
```
krmanik, line 19: I think this changes needs to push also to the repository.

tools/web/build-web-assets.sh
krmanik: The `$BUILD_DIR/web` dir created before copying files.

tools/web/build-web-assets.bat
krmanik: I have removed the `ts` build now.

krmanik:
Fixing now
```
> Task :rsdroid:buildAnkiWebAssets
D:\a\Anki-Android-Backend\Anki-Android-Backend>bazel --version   1>nul 2>&1  && (echo "bazel command found!" )  || (
echo "bazel: command not found. Please install Bazelisk. Your distro may have it,"  
 echo "or you can fetch the binary from https://github.com/bazelbuild/bazelisk/releases"  
 echo "and add it to path."  
 exit 1 
) 
"bazel command found!"
```

tools/web/build-web-assets.bat
```diff
@@ -0,0 +1,33 @@
...
+)
+
+set SRC_DIR=%cd%
+set ANKI_SRC=%SRC_DIR%\rslib-bridge\anki
+set BUILD_DIR=%SRC_DIR%\rsdroid\build\generated\anki_artifacts
+
+:: generate buildinfo.txt file
+cd %ANKI_SRC%
+bazel build --symlink_prefix=%temp%\.bazel\ -c opt buildinfo.txt
+cp %temp%\.bazel\bin\buildinfo.txt %SRC_DIR%\rslib-bridge
```
dae, line 20: In my original attempt, I changed cargo.toml to point to the generated file, instead of copying it into the tree:

https://github.com/ankidroid/Anki-Android-Backend/pull/202/commits/60a4233beb21de74f12c672291486898fb11c777#

The build failed on a clean build, I suspect because the assets weren't being built early enough.

It's possible that issue still exists - we don't know if the build process is using the copied version of the file, or the original version that's committed into the repo. For that reason, I think it would be safer to point to the generated location, and remove the currently committed file. But if it breaks on a clean build, it will also require figuring out how to build the assets earlier (I wasn't sure how). If you're not sure either, one option may be to have the user manually run a command to build the assets, before starting the rest of the build, until a gradle expert can improve things.

dae:
Thanks for all your work on this Mani, the caching looks like a nice improvement.

tools/web/build-web-assets.bat
krmanik: I think following will work, if work then the `buildAnkiWebAssets` can be removed.
```
./tools/web/build-web-assets.sh && ./gradlew clean assembleRelease -DtestBuildType=release
```

I have updated the script to copy the buildinfo to build dir. I think buildinfo.txt file is generated before build process. The local build on linux generated the web assets and copied to `web` dir.

I have changed buildinfo.txt to build dir.
```
BUILDINFO = { value = "rsdroid/build/generated/anki_artifacts/buildinfo.txt", relative = true }
```
Now, there is no need to commit the buildinfo file now.

krmanik:
For linux the build is working as expected but failing on macOS and windows.
<details>
  <summary>Linux actions result</summary>

```
> Task :clean UP-TO-DATE
> Task :rsdroid:clean UP-TO-DATE
> Task :rsdroid-instrumented:clean UP-TO-DATE
> Task :rsdroid-testing:clean UP-TO-DATE
> Task :rsdroid:buildAnkiWebAssets
2022/07/06 13:40:55 Downloading https://releases.bazel.build/5.1.1/release/bazel-5.1.1-linux-x86_64...
Starting local Bazel server and connecting to it...
Loading: 
Loading: 0 packages loaded
Loading: 0 packages loaded
Analyzing: target //:buildinfo.txt (1 packages loaded, 0 targets configured)
Analyzing: target //:buildinfo.txt (54 packages loaded, 186 targets configured)
INFO: Analyzed target //:buildinfo.txt ([57](https://github.com/ankidroid/Anki-Android-Backend/runs/7215801459?check_suite_focus=true#step:16:58) packages loaded, 217 targets configured).
INFO: Found 1 target...
[0 / 4] [Prepa] Creating source manifest for //tools:buildinfo [for host]
Target //:buildinfo.txt up-to-date:
  /tmp/.bazel/bin/buildinfo.txt
INFO: Elapsed time: 5.464s, Critical Path: 0.22s
INFO: 5 processes: 4 internal, 1 linux-sandbox.
INFO: Build completed successfully, 5 total actions
INFO: Build completed successfully, 5 total actions
Loading: 
Loading: 0 packages loaded
INFO: Build option --compilation_mode has changed, discarding analysis cache.
Analyzing: target //qt/aqt/data/web/pages:pages (1 packages loaded, 0 targets configured)
Analyzing: target //qt/aqt/data/web/pages:pages (56 packages loaded, 545 targets configured)
Analyzing: target //qt/aqt/data/web/pages:pages (79 packages loaded, 8533 targets configured)
Analyzing: target //qt/aqt/data/web/pages:pages (166 packages loaded, 13657 targets configured)
Analyzing: target //qt/aqt/data/web/pages:pages (316 packages loaded, 18449 targets configured)
Analyzing: target //qt/aqt/data/web/pages:pages (346 packages loaded, 21065 targets configured)
INFO: Analyzed target //qt/aqt/data/web/pages:pages (361 packages loaded, 21555 targets configured).
INFO: Found 1 target...
[0 / 1] [Prepa] BazelWorkspaceStatusAction stable-status.txt
[64 / 179] checking cached actions
Target //qt/aqt/data/web/pages:pages up-to-date:
  /tmp/.bazel/bin/qt/aqt/data/web/pages/graphs-base.css
  /tmp/.bazel/bin/qt/aqt/data/web/pages/graphs.css
  /tmp/.bazel/bin/qt/aqt/data/web/pages/graphs.html
  /tmp/.bazel/bin/qt/aqt/data/web/pages/graphs.js
  /tmp/.bazel/bin/qt/aqt/data/web/pages/graphs.js.map
  /tmp/.bazel/bin/qt/aqt/data/web/pages/congrats-base.css
  /tmp/.bazel/bin/qt/aqt/data/web/pages/congrats.css
  /tmp/.bazel/bin/qt/aqt/data/web/pages/congrats.html
  /tmp/.bazel/bin/qt/aqt/data/web/pages/congrats.js
  /tmp/.bazel/bin/qt/aqt/data/web/pages/congrats.js.map
  /tmp/.bazel/bin/qt/aqt/data/web/pages/deck-options-base.css
  /tmp/.bazel/bin/qt/aqt/data/web/pages/deck-options.css
  /tmp/.bazel/bin/qt/aqt/data/web/pages/deck-options.html
  /tmp/.bazel/bin/qt/aqt/data/web/pages/deck-options.js
  /tmp/.bazel/bin/qt/aqt/data/web/pages/deck-options.js.map
  /tmp/.bazel/bin/qt/aqt/data/web/pages/change-notetype-base.css
  /tmp/.bazel/bin/qt/aqt/data/web/pages/change-notetype.css
  /tmp/.bazel/bin/qt/aqt/data/web/pages/change-notetype.html
  /tmp/.bazel/bin/qt/aqt/data/web/pages/change-notetype.js
  /tmp/.bazel/bin/qt/aqt/data/web/pages/change-notetype.js.map
  /tmp/.bazel/bin/qt/aqt/data/web/pages/card-info-base.css
  /tmp/.bazel/bin/qt/aqt/data/web/pages/card-info.css
  /tmp/.bazel/bin/qt/aqt/data/web/pages/card-info.html
  /tmp/.bazel/bin/qt/aqt/data/web/pages/card-info.js
  /tmp/.bazel/bin/qt/aqt/data/web/pages/card-info.js.map
  /tmp/.bazel/bin/qt/aqt/data/web/pages/fields-base.css
  /tmp/.bazel/bin/qt/aqt/data/web/pages/fields.css
  /tmp/.bazel/bin/qt/aqt/data/web/pages/fields.html
  /tmp/.bazel/bin/qt/aqt/data/web/pages/fields.js
  /tmp/.bazel/bin/qt/aqt/data/web/pages/fields.js.map
  /tmp/.bazel/bin/qt/aqt/data/web/pages/import-csv-base.css
  /tmp/.bazel/bin/qt/aqt/data/web/pages/import-csv.css
  /tmp/.bazel/bin/qt/aqt/data/web/pages/import-csv.html
  /tmp/.bazel/bin/qt/aqt/data/web/pages/import-csv.js
  /tmp/.bazel/bin/qt/aqt/data/web/pages/import-csv.js.map
INFO: Elapsed time: 14.442s, Critical Path: 1.34s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
Loading: 
Loading: 0 packages loaded
Analyzing: target //ts/reviewer:reviewer_extras_bundle (1 packages loaded, 0 targets configured)
INFO: Analyzed target //ts/reviewer:reviewer_extras_bundle (3 packages loaded, 167 targets configured).
INFO: Found 1 target...
[0 / 1] [Prepa] BazelWorkspaceStatusAction stable-status.txt
[19 / 21] Compiling TypeScript project //ts/reviewer:reviewer_ts [tsc -p ts/reviewer/tsconfig.json]; 1s linux-sandbox
[20 / 21] [Prepa] Bundling Javascript ts/reviewer/reviewer_extras.ts [esbuild]
Target //ts/reviewer:reviewer_extras_bundle up-to-date:
  /tmp/.bazel/bin/ts/reviewer/reviewer_extras_bundle.js
  /tmp/.bazel/bin/ts/reviewer/reviewer_extras_bundle.js.map
  /tmp/.bazel/bin/ts/reviewer/reviewer_extras_bundle_metadata.json
INFO: Elapsed time: 4.699s, Critical Path: 3.[60](https://github.com/ankidroid/Anki-Android-Backend/runs/7215801459?check_suite_focus=true#step:16:61)s
INFO: 21 processes: 12 internal, 9 linux-sandbox.
INFO: Build completed successfully, 21 total actions
INFO: Build completed successfully, 21 total actions
> Task :generateLinkerWrapper
> Task :rsdroid:cargoBuildArm
```  

</details>

rsdroid/build.gradle
```diff
@@ -111,18 +109,25 @@ task generateTranslations(type: Exec) {
...
-
 preBuild.dependsOn "generateTranslations"
 
 Boolean wantAllPlatforms = System.getenv("CURRENT_ONLY") != "true"
 
-if (wantAllPlatforms) {
-    preBuild.dependsOn "cargoBuild"
-} else {
-    preBuild.dependsOn "cargoBuildX86_64"
+task cargoBuildTask(type: Exec) {
```
dae, line 116: With the following change, it builds correctly for me on a Linux machine:

task cargoBuildTask {

dae: This also seems to work:

```diff
diff --git a/rsdroid/build.gradle b/rsdroid/build.gradle
index 71a4a00bb..2e9130643 100644
--- a/rsdroid/build.gradle
+++ b/rsdroid/build.gradle
@@ -113,15 +113,12 @@ preBuild.dependsOn "generateTranslations"
 
 Boolean wantAllPlatforms = System.getenv("CURRENT_ONLY") != "true"
 
-task cargoBuildTask(type: Exec) {
+task cargoBuildTask {
     if (wantAllPlatforms) {
         preBuild.dependsOn "cargoBuild"
     } else {
         preBuild.dependsOn "cargoBuildX86_64"
     }
-}
-
-cargoBuildTask.configure {
     mustRunAfter "buildAnkiWebAssets"
 }
 
```

dae: Also rslib-bridge/buildinfo.txt looks like it can be removed

krmanik: It seems that build pass on linux and macOS. But I have change the script to `powershell` for windows but still not passing. For windows `buildinfo.txt` and `reviewer_extras_bundle.js` successfully build but for `pages` have following errors.

```
[0 / 1] [Prepa] BazelWorkspaceStatusAction stable-status.txt
ERROR: D:/a/anki-android-backend/anki-android-backend/rslib-bridge/anki/sass/bootstrap/BUILD.bazel:5:14: CopyFile sass/bootstrap/scss/_accordion.scss failed: (Exit 1): bash failed: error executing command bash bazel-out/x64_windows-fastbuild/bin/sass/bootstrap/sass-sources-cp.sh
Target //qt/aqt/data/web/pages:pages failed to build
Use --verbose_failures to see the command lines of failed build steps.
ERROR: D:/a/anki-android-backend/anki-android-backend/rslib-bridge/anki/ts/graphs/BUILD.bazel:11:13 SassCompiler ts/graphs/graphs-base.css failed: (Exit 1): bash failed: error executing command bash bazel-out/x64_windows-fastbuild/bin/sass/bootstrap/sass-sources-cp.sh
INFO: Elapsed time: 9.023s, Critical Path: 0.34s
INFO: 4 processes: 4 internal.
FAILED: Build did NOT complete successfully
Windows Subsystem for Linux has no installed distributions.
Distributions can be installed by visiting the Microsoft Store:
https://aka.ms/wslstore
FAILED: Build did NOT complete successfully
Copy-Item : Cannot find path 'C:\Users\runneradmin\AppData\Local\Temp\.bazel\bin\qt\aqt\data\web\pages' because it 
does not exist.
At D:\a\Anki-Android-Backend\Anki-Android-Backend\tools\web\build-web-assets.ps1:17 char:1
```

krmanik: I have added the workaround.
```
echo "Using git bash for bazel build"
$env:BAZEL_SH = "c:\Program Files\Git\bin\bash.exe"
```

krmanik: I am not able to get pass this errors
```
[0 / 1] [Prepa] BazelWorkspaceStatusAction stable-status.txt
ERROR: D:/a/anki-android-backend/anki-android-backend/rslib-bridge/anki/sass/bootstrap/BUILD.bazel:5:14: CopyFile sass/bootstrap/scss/_accordion.scss failed: (Exit 1): bash failed: error executing command bash bazel-out/x64_windows-fastbuild/bin/sass/bootstrap/sass-sources-cp.sh
Windows Subsystem for Linux has no installed distributions.
Distributions can be installed by visiting the Microsoft Store:
https://aka.ms/wslstore
Target //qt/aqt/data/web/pages:pages failed to build
```

I have tried following
- `--shell_executable`
```
bazel build --symlink_prefix=$TEMP_DIR/.bazel/ -c opt buildinfo.txt --verbose_failures --shell_executable="C:\\Program Files\\Git\\usr\\bin\\bash.exe"
```

- Setting path
```
BAZEL_SH="C:\\Program Files\\Git\\usr\\bin\\bash.exe"
```

- Used powershell to set alias
```
Set-Alias -Name bash -Value "C:\Program Files\Git\usr\bin\bash.exe"
```

- Changed the gradle to use git bash
```diff
task buildAnkiWebAssets(type: Exec) {
    workingDir "$rootDir"
    String toolPath = 'tools/web/build-web-assets.sh'
    if (System.getProperty('os.name').toLowerCase(Locale.ROOT).contains('windows')) {
-        commandLine 'cmd', '/c', 'Powershell  -File', toolPath
+        commandLine 'C:\\Program Files\\Git\\usr\\bin\\bash.exe', toolPath
    } else {
        commandLine 'sh', '-c', toolPath
    }
}
```

The `buildinfo.txt` and `reviewer_extras_bundle.js` file generated but `pages` build not passing on windows.

krmanik: > * Have you confirmed that the env vars you're setting in windows-build.yml are actually still set when the ps1 file is run? Maybe they could be set directly in the ps1 file?

I have tried this, added env to ps1 file, but still errors on bash.

> * When you tried --shell_executable, did you modify all bazel build invocations? what about if you use c:/msys64/... (forward slash, msys bash instead of git)?

Also changed the bash to  c:/msys64/ but not passing. 

Trying adding vars in gradle script before ps1 execution.

krmanik: Or can we use https://github.com/Vampire/setup-wsl to setup wsl and build/copy the assets ?

dae: @mikehardy @david-allison I'm not sure how this can be fixed. I see two paths forward:

- We leave the Github Windows action as is - it's not correctly copying the files, but it runs to completion. Local users could presumably get things working by the hack I mentioned in my reply above
- We generate the assets using WSL, then run the rest of the build normally. This should fix building with GitHub actions, but will mean users who want to build locally will now need to have WSL set up.

Since the official releases are cross-compiled and don't actually need the Windows build to run correctly, the first approach seems like less work - WDYT?

krmanik:
> That latest change looks promising. The new error might just be a flake - you could try running that bazel build line a few times in a row.

Added in ps1
```
for (($i = 0); $i -lt 5; $i++)
{
    bazel build pages -k --shell_executable="C:/msys64/usr/bin/bash.exe"
}
```

dae:
No luck :-( Some Googling seems to indicate TypeScript can give that error when using symlinks, but I guess we're not using them there? Maybe it's some other path-related issue. That path is quite long - I wonder if you copied rslib-anki/anki to d:\a\anki and built from there, if it would help at all?

Another possible thing you could try is the -win.sh file now that you've fixed the bash issue

krmanik:
```
2 [main] cp (1552) C:\msys64\usr\bin\cp.exe: *** fatal error - cygheap base mismatch detected - 0x18034C408/0x18034B408.
This problem is probably due to using incompatible versions of the cygwin DLL.
Search for cygwin1.dll using the Windows Start->Find/Search facility
and delete all but the most recent version.  The most recent version *should*
reside in x:\cygwin\bin, where 'x' is the drive on which you have
installed the cygwin distribution.  Rebooting is also suggested if you
are unable to find another cygwin DLL.
```

Trying now `git bash` instead of `msys2 bash`

mikehardy:
A vague note but

>  That path is quite long

This is definitely a problem with the windows build of some core react-native components now and they do some path-shortening hacks. Is that what's happening here? No idea but good to note that it is a possibility

krmanik:
I have tried this for finding path related issues on temp repository and test ps1 file but errors still same.
```
$SRC_DIR = (Get-Item D:/a/).FullName
Set-Location -Path $SRC_DIR

git clone https://github.com/ankitects/anki.git

Set-Location -Path anki/qt/aqt/data/web/pages
bazel build pages
```

Errors
```
[326 / 378] Action ts/lib/backend_proto.d.ts; 6s local ... (2 actions running)
ERROR: D:/a/anki/ts/sveltelib/BUILD.bazel:5:11: Compiling TypeScript project //ts/sveltelib:sveltelib [tsc -p ts/sveltelib/tsconfig.json] failed: (Exit 1): tsc.bat failed: error executing command bazel-out\host\bin\external\npm\typescript\bin\tsc.bat --project bazel-out/x64_windows-fastbuild/bin/ts/sveltelib/tsconfig.json --outDir bazel-out/x64_windows-fastbuild/bin/ts/sveltelib --rootDir ... (remaining 6 arguments skipped)
D:\a\anki\node_modules\typescript\lib\tsc.js:1695
        return str.lastIndexOf(prefix, 0) === 0;
                   ^
TypeError: Cannot read properties of undefined (reading 'lastIndexOf')
    at Object.startsWith
```

I think we should continue on windows without building `pages`. The buildinfo.txt build works so we use this only.

krmanik:
I think the build passing when using this the `--output_user_root`.
```
bazel --output_user_root=D:/a/_bzl build pages
``` 

```
[0 / 39] [Prepa] BazelWorkspaceStatusAction stable-status.txt
[74 / 238] Compiling util/process_wrapper/process_wrapper.cc [for host]; 1s local ... (2 actions, 1 running)
[171 / 302] Compiling Rust rlib cargo_build_script_output_parser (1 files) [for host]; 0s local ... (2 actions, 1 running)
[304 / 374] Compiling Rust rlib syn v1.0.88 (90 files) [for host]; 15s local ... (2 actions, 1 running)
[307 / 374] Compiling Rust rlib syn v1.0.88 (90 files) [for host]; 51s local ... (2 actions running)
[315 / 374] Compiling Rust proc-macro thiserror_impl v1.0.30 (8 files) [for host]; 31s local ... (2 actions running)
[331 / 377] Action ts/lib/backend_proto.d.ts; 4s local ... (2 actions, 1 running)
[545 / 620] Compiling TypeScript project //ts/components:components [tsc -p ts/components/tsconfig.json]; 4s local ... (2 actions, 1 running)
[588 / 630] Compiling Svelte ts/fields:svelte; 1s worker ... (2 actions, 1 running)
```

dae:
Great work in figuring that out Mani. Assuming you've tested that the pages can be successfully loaded on the AnkiDroid end with this build, it's a :+1: from me. Thank you for all of your hard work!

krmanik:
I have tested `rsdroid-release.aar` file by publishing from GitHub using release actions and on AnkiDroid the pages loads as expected. 

On windows machine it needs `Build Tools for Visual Studio 2019 (version 16.11)` to build the pages using the docs [windows.md](https://github.com/ankitects/anki/blob/main/docs/windows.md) (_I haven't tried this_).

dae:
Could we include the license files too? https://github.com/ankidroid/Anki-Android/issues/11825#issuecomment-1179828583

dae:
Re docs, how about:

```diff
diff --git a/docs/TESTING.md b/docs/TESTING.md
index 9a95dcfc6..1213ee562 100644
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -8,7 +8,11 @@ so far.

 ## Setup

-Make sure you can build AnkiDroid first.
+- Make sure you can build AnkiDroid first.
+
+- Follow the initial setup instructions in the computer version
+  (rslib-bridge/anki/docs/{windows,mac,linux}.md) to install prerequisites
+  like Bazelisk and the build tools for your platform.

 Install NDK:
```

krmanik:
> Could we include the license files too? [ankidroid/Anki-Android#11825 (comment)](https://github.com/ankidroid/Anki-Android/issues/11825#issuecomment-1179828583)

The `licenses.json` from `cargo` and `ts` included as ` licenses-cargo.json` and ` licenses-ts.json` in web dir.

> Re docs, how about:
> 

Thanks, I have updated the docs.

.github/workflows/robolectric_build.yml
```diff
@@ -173,10 +200,38 @@ jobs:
           pip3 install --upgrade protobuf stringcase
           python3 .github/scripts/protoc_gen_deps.py
 
+      - name: Setup bazel
```
mikehardy, line 203: None of the bazelisk install seems necessary, all our virtual environments don't just include it, they include the current (1.12.0) version. Here we are manually installing it, but partially as v1.11.0 and partially v1.10.1

https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md#tools
https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2204-Readme.md#tools
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-12-Readme.md#utilities

Is there a reason why we are either not compatible with current version, or cannot use the pre-installed version?

rsdroid/build.gradle
```diff
@@ -45,10 +45,23 @@ android {
...
         main {
             kotlin.srcDirs += "build/generated/source/fluent"
+            resources {
+                srcDirs "src/main/resources", "build/generated/anki_artifacts"
+            }
         }
     }
 }
 
+task buildAnkiWebAssets(type: Exec) {
```
mikehardy, line 55: I don't understand why the custom task is duplicated - it seems like it could be in the top-level build.gradle, or in a separate gradle file that rsdroid-testing and rsdroid both include?

mikehardy (changes requested):
Getting it to the point where it is working is fantastic (and I think it is there right?) - that's the hard part
My comments are now about editing it back down to be minimal by removing any duplication that can be removed, if possible

.github/workflows/robolectric_build.yml
krmanik: Thanks, I didn't know about it. I have removed the bazelisk from yml file.

rsdroid/build.gradle
krmanik: I moved the task another file.

.github/workflows/linux_build.yml
```diff
@@ -70,6 +70,27 @@ jobs:
...
+      - name: Node module cache
+        id: node-module-cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            rslib-bridge/anki/node_modules
+          key: ${{ runner.os }}-v1-node-module-cache-${{ hashFiles('rslib-bridge/anki/package.json') }}
+          restore-keys: |
+            ${{ runner.os }}-v1-node-module-cache-
+            ${{ runner.os }}-v1-node-module-cache
```
mikehardy, line 92: These two lines look odd to me (here and in all the .yml files) - is this intentional? Or should they all be

```suggestion
            ${{ runner.os }}-v1-node-module-cache
```

mikehardy: Also just as a side note, I do a lot of github actions .yml work and this is not something we can change here easily - but I wish these blocks were not duplicated everywhere. I find code sharing / code abstraction in the github actions yml blocks to be kind of annoying and I haven't figured out a direction I want to go in yet.

Please don't take any action here - I want this PR to go through really soon :-) - but I know you are doing a lot of yml work now too, if you can think of a good + easy way to de-duplicate these "cache the same thing in all our workflows, hashing the same file, with the same keys" stuff, that would be great as a follow-up. Not a priority though, just a thought. Cheers

.github/workflows/windows_build.yml
```diff
@@ -92,6 +92,27 @@ jobs:
             ${{ runner.os }}-rust-v1-assembleRelease
             ${{ runner.os }}-rust-v1
 
+      - name: Bazel cache
+        id: bazel-cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            C:\Users\runneradmin\_bazel_runneradmin
```
mikehardy, line 100: This path appears to vary between mac, linux, and windows. It makes sense on windows (paths always vary there) but I don't understand the linux / mac difference. Can you just add a comment on the mac one explaining why the bazel cache is in a different place? That would be sufficient to merge. For full bonus points - not required - if it's easy to do a platform-conditional block where you set a workflow variable to the path (with a comment on why each platform has the path it does) that would be great, but when I have not done that for a while personally I get stuck doing variable setting / variable value de-referencing and it takes a while. When I have a test rig all set up or I've done it recently it's fast but you may not have that right now. So don't spend much time on it if isn't fast for you - a comment is fine.

Thanks Mani

mikehardy (changes requested):
Mostly one question on the cache restore keys

One note on the cache paths for bazel action, and a simple comment could get this merged, or variable interpolation if that's fast for you (just a comment if not...)

rsdroid/build.gradle
david-allison: (open to discussion).

I'd go all-in with WSL2 on Windows, with the plan to move to native if/when we find a maintainer who's using Windows. As long as CI passes then users have a path to installation

Building Rust targeting Android on Windows wasn't well maintained. My first issue with this project was OpenSSL. I expect there to be significant bitrot here as it's an uncommon combination of targets, and we don't have currently have a maintainer who has the capacity to fix the downstream issues (I barely have the capacity for AnkiDroid currently, and would be starting from 0 learning Perl to get things working here).

Cross-compiling for Windows on WSL is very simple in comparison

There's a significant time cost in reducing the time to setup this library, and (sadly) I don't think it will pay off in the case of Windows-based maintainers. WSL is the easier option until someone's willing to throw their hat into the ring and maintain a native build.

* The backend is vital for the project, so people will expend effort when they need to
* The backend is a library, and not UI-based, we're not going to get the level of drive-by contributors. People here will naturally be more invested

mikehardy: I would be okay with a WSL build if we did in fact go all in on it.
WSL used to be sort of a pain with other emulation systems (like: the android emulator) but they play nicely now, and it's all free and staying up to date etc, I have no issue with the idea of WSL as a foundational layer.

If this sounds different than my previous comment I guess the difference is one of commitment. If it's *all* WSL, okay then. You set the expectation and people just work with WSL as a bedrock dependency.

The problem I had in another project was a part-WSL / part-native thing, and that was terrible (it was my attempted solution, and it was terrible, I can say that ;-) - live and learn ). Just no half-measures

.github/workflows/linux_build.yml
krmanik: I have reduce the duplication. I concatenated the hash of `defs.bzl` and `package.json`. 
The cache needs to be included in each yml file, so I think, may be no further de-duplication can be done.

.github/workflows/linux_build.yml
```diff
@@ -70,26 +70,15 @@ jobs:
...
-            ${{ runner.os }}-v1-bazel-cache
-
-      - name: Node module cache
-        id: node-module-cache
-        uses: actions/cache@v2
-        with:
-          path: |
             rslib-bridge/anki/node_modules
-          key: ${{ runner.os }}-v1-node-module-cache-${{ hashFiles('rslib-bridge/anki/package.json') }}
+          key: ${{ runner.os }}-v1-node-bazel-cache-${{ hashFiles('rslib-bridge/anki/package.json') }}-${{ hashFiles('rslib-bridge/anki/defs.bzl') }}
```
mikehardy, line 79: interesting idea - this seems like it might invalidate the caches a little bit more frequently then if they are split, but how much more frequently really? And I like how it kills half the caching. I like it

.github/workflows/windows_build.yml
krmanik: I think the url should be added as comment, because there no reason provided on docs for varying paths.
https://docs.bazel.build/versions/main/output_directories.html#:~:text=The%20outputRoot%20directory%20defaults%20to,with%20the%20FOLDERID_Profile%20flag%20set.

> The outputRoot directory defaults to ~/.cache/bazel on Linux, /private/var/tmp on macOS, and on Windows it defaults to %HOME% if set, else %USERPROFILE% if set, else the result of calling SHGetKnownFolderPath() with the FOLDERID_Profile flag set. If the environment variable $TEST_TMPDIR is set, as in a test of bazel itself, then that value overrides the default.

mikehardy: excellent - works for me - go for it with just the comment then we can land this!

krmanik: I have added the comments.

mikehardy (approved):
This is a big advance - thank you!

.github/workflows/robolectric_build.yml
```diff
@@ -228,4 +228,10 @@ jobs:
 
       - name: Run Tests (win)
         if: matrix.os == 'windows-latest'
-        run: ./gradlew rsdroid:test -x jar -x cargoBuildArm -x cargoBuildX86 -x cargoBuildArm64 -x cargoBuildX86_64
+        run: | 
+          echo "Removing Bash from WSL"
```
david-allison, line 195: Could this be updated to explain why we're using msys2 instead (maybe do it on the shell script?)

.github/workflows/windows_build.yml
```diff
@@ -134,4 +138,9 @@ jobs:
           $Env:ANDROID_NDK_PATH =  "C:\ProgramFiles\Android\android-sdk\ndk-bundle"
           $Env:ANDROID_SDK_ROOT =  "C:\ProgramFiles\Android\android-sdk"
           $env:ANDROID_HOME      
+          echo "Removing Bash from WSL"
```
david-allison, line 121: ditto

.github/workflows/windows_pure_build.yml
```diff
@@ -95,4 +99,10 @@ jobs:
             ${{ runner.os }}-v1-node-module-cache
 
       - name: Build
-        run: ./gradlew clean assembleRelease -DtestBuildType=release
+        run: | 
+          echo "Removing Bash from WSL"
```
david-allison, line 83: ditto

tools/web/build-web-assets.ps1
```diff
@@ -1,18 +1,21 @@
+Set-Alias -Name bash -Value "C:/msys64/usr/bin/bash.exe"
+$env:BAZEL_SH = "C:/msys64/usr/bin/bash.exe"
```
david-allison, line 2: follow up to make these paths variable. Maybe fail if we're not using msys2 (if it's required)

build-web-assets.gradle
```diff
@@ -0,0 +1,9 @@
+task buildAnkiWebAssets(type: Exec) {
+    workingDir "$rootDir"
+    String toolPath = System.getProperty('os.name').toLowerCase(Locale.ROOT).contains('windows') ? 'tools/web/build-web-assets.ps1' : 'tools/web/build-web-assets.sh'
```
david-allison, line 3: follow-up: we should abstract out checking the OS to a method

david-allison (commented):
Nothing major. 

I'm out of my comfort zone here and trust dae/Mani/Mike's judgment 100% on this

.github/workflows/robolectric_build.yml
krmanik: I will move it to shell script.

rsdroid/build.gradle
dae: Mani was able to fix the build by removing the bash stub in the Windows folder, so there's no pressing need to switch to WSL at the moment.

Perl should no longer be an issue, as the backend is being built with https://github.com/rustls/rustls instead of OpenSSL

mikehardy: Okay - resolving (at least for now - we can always revisit) as "staying native on windows" for the moment

.github/workflows/robolectric_build.yml
mikehardy: for now just the comment works for me - I'd really like to get this in as it's blocking a stack of work built on top of a needed release of Anki-Android-Backend and this solution was arrived at after lots of trial and error

krmanik: Just moved the env path setup to shell script, hope it should work.

krmanik: I have reverted the commit and added the comments. I will try it in follow up PR and may be try to integrate WSL.
```
+ Move-Item -Path "C:/WINDOWS/system32/bash.exe" -Destination "C:/WINDO ...
Removing Bash from WSL
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : PermissionDenied: (C:\WINDOWS\system32\bash.exe:FileInfo) [Move-Item], UnauthorizedAcces 
   sException
    Directory: D:\a\Anki-Android-Backend\Anki-Android-Backend\rsdroid\build\generated\anki_artifacts
    + FullyQualifiedErrorId : MoveFileInfoItemUnauthorizedAccessError,Microsoft.PowerShell.Commands.MoveItemCommand
```

mikehardy: okay now there is a link to the upstream issue and it contains essentially this workaround (plus a link back to us from the related work here ;-) ) - good enough for future explorers I hope

mikehardy (approved):
Thanks Mani! I'm going to merge this to unblock things in my last time available for a while, we can always do fixups as needed / if needed

PR #218, opened 2022-06-30
Allow user to cancel operations like syncs and exports

dae:
No hurry for a new release


mikehardy (approved):
LGTM - this is a pretty heavy release process, maybe a 1-a-week max type thing or something? Or wait and have it be "demand pull", like, when there is something in Anki-Android that is important/useful for whatever reason and is fully blocked?

If I understand correctly this is more "nice to have" for user experience in non-legacy schema so it's critical at all to release, is that correct?

dae:
Yeah, wouldn't expect you to be updating this frequently; let's wait until there's a blocker or enough time has passed.

PR #225, opened 2022-07-16
Update README.md

yyy153:
Testing library to allow the above to be usable under Robolectric -> Test library to allow the above to be usable under Robolectric


mikehardy (changes requested):
I don't think this is an improvement

PR #227, opened 2022-07-24
Fix successive invocations not updating buildinfo

dae:
Doesn't affect CI; not urgent. Also fail the script if a subcommand
fails.

PS: the 1.15 release is missing a tag

### DIFF
--- a/tools/web/build-web-assets.sh
+++ b/tools/web/build-web-assets.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 # build web assets with anki submodule in rslib-bridge
 
+set -e
+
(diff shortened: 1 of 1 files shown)


mikehardy:
> PS: the 1.15 release is missing a tag

Thanks for the reminder - helps me keep things together at the moment

Did that whole release (reviews, actions etc) mobile - availability is a bit like that at the moment. Got a moment on a real machine and tagged it up https://github.com/ankidroid/Anki-Android-Backend/releases/tag/0.1.15-anki2.1.54

mikehardy (approved):

PR #228, opened 2022-07-24
Declare backend methods with @CheckResult

dae:
This currently doesn't seem to work - when viewing the decompiled class file in the AnkiDroid repo or hovering over a backend method, it shows the annotation has been included, but for some reason it is ineffective: it does not trigger a lint error, and the IDE does not show a warning. Any ideas?


dae:
This has code rot and is not super important, so I'm just going to close it.

Issue #235, opened 2022-09-19
Determine if/how to update upstream translations in Anki-Android-Backend


mikehardy:
We need to work on frequency/freshness of the Pontoon translations which are at this point pretty static (we have not ingested changes from that system in months, and...have no plan to do so right now which is not good, I want a plan we just don't have one). I'm not sure if we can ingest just translations easily or if we must necessarily update the whole backend or not, the question bears investigating

_Originally posted by @mikehardy in https://github.com/ankidroid/Anki-Android/issues/12448#issuecomment-1251267309_

dae:
Each desktop release pins the strings at a particular point, since newer code may depend on new strings, and obsolete strings may be dropped. You thus can't update the translations separately from the anki module without risk of things falling out of sync. 2.1.55 is hopefully not more than a month away, and when it comes out and a new backend release is made, all the accumulated contributions will be in it. AD doesn't necessarily have to stick to stable releases either - it can update the backend before then if it wishes.

mikehardy:
Sounds like the same process we have - code and strings need to match (for same reason - string add/drops).
So the process for ingestion of new strings will be to "re-sync" the backend to some specific commit - either a stable tag or some interesting commit, no matter from a sync perspective, it's a commit hash at the end.

A finer question: is there any risk in any specific commits? For example, we typically commit a PR that does string add/drops *then* we do an i18n sync with crowdin that populates the other languages and commit that. So from an i18n perspective there is a brief moment where strings are out of sync for non-english locales. I don't think that actually has a risk of crashing in Android (it should fall back to the default locale for missing strings?) but maybe it's different for backend?

That would inform whether commits to sync backend need any care when they are chosen or really if any commit would work (assuming it was a "good" commit to choose from a functionality/error perspective)

dae:
Generally any commit on main is safe, though there have been a few instances in the past where a bug was introduced and then fixed in a follow-up commit after it was discovered.

Issue #213, opened 2022-06-28
working directory is incorrect for `protoc-gen.sh` when run from Android Studio


david-allison:
PWD = `/Users/davidallison/.gradle/daemon/6.8.3`

```
Execution failed for task ':rsdroid:generateDebugProto'.
> protoc: stdout: . stderr: PWD = /Users/davidallison/.gradle/daemon/6.8.3
  /Users/davidallison/StudioProjects/Anki-Android-Backend/tools/protoc-gen/protoc-gen.sh: line 5: tools/setup-python: No such file or directory
  /Users/davidallison/StudioProjects/Anki-Android-Backend/tools/protoc-gen/protoc-gen.sh: line 6: ./tools/protoc-gen/protoc-gen.py: No such file or directory
  --anki_out: protoc-gen-anki: Plugin failed with status code 127.
```

Issue #211, opened 2022-06-27
new CURRENT_ONLY build/test performance option assumes x86-64 / likely fails on M1 mac


mikehardy:
shouldn't this determine the correct platform? Seems like cargoBuildX86_64 is an assumption as what the opposite of wantAllPlatforms is. David has an M1 now, for instance. That may be fancy so perhaps a middle-ground would be to failfast if CURRENT_ONLY == true and arch != x86-64 (or linux/x86-64 if it's that specific)

_Originally posted by @mikehardy in https://github.com/ankidroid/Anki-Android-Backend/pull/202#discussion_r906897699_

mikehardy:
Got a timeslice on the M1 we have in the house and for `./test-current.sh` the appropriate thing is preBuildMac apparently:

```
Execution failed for task ':rsdroid-testing:preBuildMac'.
```

because of

```

  |
  = note: the `x86_64-apple-darwin` target may not be installed
  = help: consider downloading the target with `rustup target add x86_64-apple-darwin`

For more information about this error, try `rustc --explain E0463`.
error: could not compile `cfg-if` due to previous error

Caused by:
  process didn't exit successfully: `rustc --crate-name cfg_if --edition=2018 /Users/mike/.cargo/registry/src/github.com-1ecc6299db9ec823/cfg-if-1.0.0/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts --crate-type lib --emit=dep-info,metadata,link -C opt-level=3 -C embed-bitcode=no -C metadata=581f1a16a94707a6 -C extra-filename=-581f1a16a94707a6 --out-dir /Users/mike/work/Ankidroid/Anki-Android-Backend/rslib-bridge/target/x86_64-apple-darwin/release/deps --target x86_64-apple-darwin -L dependency=/Users/mike/work/Ankidroid/Anki-Android-Backend/rslib-bridge/target/x86_64-apple-darwin/release/deps -L dependency=/Users/mike/work/Ankidroid/Anki-Android-Backend/rslib-bridge/target/release/deps --cap-lints allow` (exit status: 1)
warning: build failed, waiting for other jobs to finish...
error: build failed

```

Here's the diff to PR after merging #202 

```

diff --git a/rsdroid/build.gradle b/rsdroid/build.gradle
index fd7bb40..d6236b9 100644
--- a/rsdroid/build.gradle
+++ b/rsdroid/build.gradle
[...]

Issue #197, opened 2022-06-11
doctor should offer use of one arch /  no cross compile


mikehardy:
Good feedback - I'd like the process to be as automated as possible, ...

- doctor should offer the option of using one arch / no docker (related #40 for the build itself)

_Originally posted by @mikehardy in https://github.com/ankidroid/Anki-Android-Backend/pull/194#issuecomment-1152955955_

Related #40

PR #246, opened 2022-11-29
Update to 2.1.55 and refactor the build system

dae:
The separate publish workflows have been removed; the "build all" one has the final publishing steps that should only trigger when the workflow is invoked manually.

I've left a few separate commits for reference, but recommend squashing when it comes time to merge in. Notes from the commit:

The 2.1.55 desktop release switched away from Bazel to a new build system,
which required some changes to this repo. Making the changes was a bit complicated
due to the complexity of the current build system, and I ended up shaving some yaks
to make things simpler while I was working on the changes.

Changes:

- Building for the current architecture instead of all architectures is now the
default, so getting started is easier, and a build in Android Studio no longer
requires special flags.
- The build-current.sh script has been split into build-aar and build-robo
for the Android and Robolectric parts, and can be used for both single-arch
and multi-arch builds.
- On arm64 Macs, the build scripts now create arm binaries
- In a multi-arch build, both x86 and arm64 Mac libs are built, and they're
merged into a single library. This can be done in CI, so there is no manual step
required for M1 machines anymore.
- The build now uses protobuf and python binaries/libs that the desktop build
downloads, so they don't need to be installed separately.
- The pinned Rust version and Rust targets are automatically installed as
required.
- The per-platform CI builds now build in debug mode and are faster.
- Updated the docs to explain how the NDK can be installed via Android Studio,
instead of via separate command-line tools.
- The cross/docker stuff has been stripped out, as it's of limited use as it
can't target macOS legally. Easier to use GH actions for the multi-arch
builds, and keep things simple for local development.
- Fix lint not being run in CI; caught an API 23 reference.

Bumps https://github.com/ankidroid/Anki-Android-Backend/issues/179 (builds on M1 already work, so this may be simpler than expected?)
Bumps https://github.com/ankidroid/Anki-Android-Backend/issues/174 (a bunch of the doctor stuff is obsolete; updated HOWTO.md and GH actions should be consulted)
Bumps https://github.com/ankidroid/Anki-Android-Backend/issues/27 (I recommend closing this; single-platform is the default for local builds, and
CI runners don't have any extra compute available)

Tentatively closes https://github.com/ankidroid/Anki-Android-Backend/issues/109 (didn't see the flake when I was updating the actions)

Closes https://github.com/ankidroid/Anki-Android-Backend/issues/235 (translation submodules now automatically synced with anki submodule)
Closes https://github.com/ankidroid/Anki-Android-Backend/issues/213 (path based on script now)
Closes https://github.com/ankidroid/Anki-Android-Backend/issues/211 (builds for Arm Mac on Arm Macs)
Closes https://github.com/ankidroid/Anki-Android-Backend/issues/197 (single arch is the default)
Closes https://github.com/ankidroid/Anki-Android-Backend/issues/196 (desktop venv is used)
Closes https://github.com/ankidroid/Anki-Android-Backend/issues/195 (can be done via the GUI, and does not require separate cli download)
Closes https://github.com/ankidroid/Anki-Android-Backend/issues/168 (latest Rust; easier changing via rust-toolchain.toml)
Closes https://github.com/ankidroid/Anki-Android-Backend/issues/164 (universal dylib)
Closes https://github.com/ankidroid/Anki-Android-Backend/issues/127 (no docker)
Closes https://github.com/ankidroid/Anki-Android-Backend/issues/106 (most of those scripts obsolete; some commands moved into build scripts)
Closes https://github.com/ankidroid/Anki-Android-Backend/issues/99 (no docker)
Closes https://github.com/ankidroid/Anki-Android-Backend/issues/98 (no docker)
Closes https://github.com/ankidroid/Anki-Android-Backend/issues/97 (submodule automatically updated)
Closes https://github.com/ankidroid/Anki-Android-Backend/issues/96 (build will fail if commit unavailable)
Closes https://github.com/ankidroid/Anki-Android-Backend/issues/53 (no docker)
Closes https://github.com/ankidroid/Anki-Android-Backend/issues/40 (DEBUG=1 option documented)
Closes https://github.com/ankidroid/Anki-Android-Backend/issues/9 (simpler OOTB experience, and updated docs)


dae:
Workflow runs on my fork:

https://github.com/ankitects/Anki-Android-Backend/actions/runs/3569944769
https://github.com/ankitects/Anki-Android-Backend/actions/runs/3569944767

rsdroid/build.gradle
```diff
@@ -103,13 +104,18 @@ task generateTranslations(type: Exec) {
 
 preBuild.dependsOn "generateTranslations"
 
-Boolean wantAllPlatforms = System.getenv("CURRENT_ONLY") != "true"
+Boolean wantAllPlatforms = System.getenv("ALL_ARCHS") == "1"
+print("archs: ${System.getenv("ALL_ARCHS")}")
```
mikehardy, line 108: A little less terse message?

```suggestion
print("building all archs? ${System.getenv("ALL_ARCHS")}")
```

rsdroid/src/main/java/net/ankiweb/rsdroid/Backend.kt
```diff
@@ -274,9 +274,9 @@ open class Backend(val context: Context, langs: Iterable<String> = listOf("en"),
         checkMainThreadOp(query)
     }
 
-    private fun checkMainThread(func: () -> Unit) {
+    private fun runIfOnMainThread(func: () -> Unit) {
         try {
-            if (Looper.getMainLooper().isCurrentThread) {
+            if (Looper.getMainLooper().thread == Thread.currentThread()) {
```
mikehardy, line 279: Interesting - that's very slightly different - was the previous test not reliable (that is: the main looper was sometimes not the current thread?)

rslib-bridge/Cargo.toml
```diff
@@ -11,26 +9,18 @@ crate_type = ["dylib"]
...
-prost = "0.9"
+anki = { path = "../anki/rslib" }
+prost = "0.11"
 serde = "1.0.114"
 serde_json = "1.0.56"
 serde_derive = "1.0.114"
 lazy_static = "1.4.0"
 num_enum = "0.5.0"
 itertools = "0.10.0"
 lexical-core = "0.7.5"
```
mikehardy, line 20: Just curious if any of the other deps are out of date, what's the command to check ? I can look it up if you don't know immediately off-hand just figured you might

docs/HOWTO.md
```diff
@@ -0,0 +1,89 @@
+# How to build, test and use this library
+
+## Prerequisites
+
+Make sure you have Android Studio can build AnkiDroid first.
```
mikehardy, line 5: Ninja is a pre-requisite isn't it? I get this when I first naively try this PR:

```
thread 'main' panicked at 'ninja not installed: Os { code: 2, kind: NotFound, message: "No such file or directory" }', build/runner/src/build.rs:66:35
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

> Task :rsdroid:buildAnkiWebAssets FAILED

```

Linux:
> sudo apt install ninja-build

macOS:
> brew upgrade
> brew install ninja

windows: Using [choco](https://chocolatey.org/):

> choco install [ninja](https://community.chocolatey.org/packages/ninja)

build-robo.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+#
+# Builds the .jar file containing the backend library compiled for 
+# desktop, which Robolectric needs. By default only the library for
+# the current architecture is built (eg X86_64, or Arm64 on Apple M1+).
+#
+# Define ALL_ARCHS=1 to build a multi-platform bundle. This will only
+# work on macOS, as it's not possible to target macOS from other platforms.
```
mikehardy, line 8: Should do a functioning version of the below to stop the curious + naive from touching the hot stove perhaps?

```bash
if ![ "$(uname)" == "Darwin" ]  # and ALL_ARCHS = 1 ? 
```

Also just to be completely clear - cross-compiling to aarch64 works from x86_64 darwin now? If so *sweet*! Because I'm not sure GitHub will ever get to allowing M1 builds in free repos in this lifetime based on the pace of their implementation (I follow one of the related issues about same)

Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[workspace.package]
+version = "0.0.0"
+authors = ["Ankitects Pty Ltd and contributors"]
+license = "AGPL-3.0-or-later"
+rust-version = "1.64"
```
mikehardy, line 5: but in rust-toolchain.toml below it is 1.65? Do they have to be duplicated, should they be in sync, or are they separate and I'm missing something?

build-aar.bat
```diff
@@ -0,0 +1,2 @@
+rustup target add x86_64-linux-android || exit /b 1
```
mikehardy, line 1: this is just for humor, but someone will try this on windows arm and this will bite later

tools/web/build-web-assets.sh
```diff
@@ -1,32 +1,18 @@
...
 
-bazel build --symlink_prefix=/tmp/.bazel/ -c opt buildinfo.txt
-cp /tmp/.bazel/bin/buildinfo.txt $BUILD_DIR/buildinfo.txt
-
-bazel build ts/reviewer/reviewer_extras_bundle --symlink_prefix=/tmp/.bazel/
-cp /tmp/.bazel/bin/ts/reviewer/reviewer_extras_bundle.js $BUILD_DIR/web
-
-bazel build qt/aqt/data/web/pages --symlink_prefix=/tmp/.bazel/
-cp -r /tmp/.bazel/bin/qt/aqt/data/web/pages/* $BUILD_DIR/web
+./ninja extract:protoc ts:reviewer:reviewer_extras_bundle.js qt/aqt:data/web/pages
```
mikehardy, line 14: So, is ninja expected in the current working directory?

docs/HOWTO.md
mikehardy: Any reason not to put this in the README? Neither file is very big, but README doesn't reference this I don't think, and "how" is the first thing I want to see as a dev in the README after the "what"/"why"

test-current.sh
```diff
@@ -2,5 +2,5 @@
 
 set -e
 
-NO_CROSS=true CURRENT_ONLY=true ./gradlew rsdroid:test rsdroid-instrumented:connectedAndroidTest
+./gradlew rsdroid:test rsdroid-instrumented:connectedAndroidTest
```
mikehardy, line 5: Requires `./build-robo.[sh|bat]` to run first, could do a check for the needed file existing first, as output from that script in this line:

>         test -f $file && cp -v $file rsdroid-testing/assets/

? Otherwise test fails for the naive executor

mikehardy (changes requested):
This is an amazing simplification, that's a shiny smooth yak you've got there! Thank you immensely

Just to confirm, M1 robo builds fine now on a mac x86-64 free runner? If so, that's huge and yes of course, this is fantastic.

Had some thoughts on docs + some guardrails in a couple of the scripts and some other little things but in general, really cool.

Any thoughts on when 2.1.55 has a final commit hash to pin?

rsdroid/src/main/java/net/ankiweb/rsdroid/Backend.kt
dae: Lint reports isCurrentThread is API 23

rslib-bridge/Cargo.toml
dae: Those are not precise pins; the exact versions will be in Cargo.lock. `cargo update` will update to the latest compatible versions, though I don't recommend doing that until the anki submodule is updated again, as the latest chrono will spit out some deprecation warnings that haven't been addressed yet (and thus has been pinned upstream for now).

docs/HOWTO.md
dae: Ninja's mentioned a few lines below - did you perhaps read the top line as "build AnkiDroidBackend" instead of "AnkiDroid"? Will tweak the text anyway.

build-robo.sh
dae: Yep, both platforms can be built on the x86_64 runners.

Cargo.toml
dae: This specifies the minimum version required to compile; rust-toolchain pins an exact version. The pinning is useful for CI, as while newer Rust versions should compile fine, they often introduce new lints that may break CI. So basically one is the minimum version to compile, the other is the maximum that is guaranteed to lint correctly, and the one that will be used.

rslib-bridge/Cargo.toml
dae: I ended up updating the submodule to fix some lints, so you can try this now after I push the changes.

build-aar.bat
dae: I look forward to their PR :-)

tools/web/build-web-assets.sh
dae: in anki/ there's a ninja script that does some extra setup work before eventually invoking ninja

dae:
> Any thoughts on when 2.1.55 has a final commit hash to pin?

Hopefully within a few weeks - there are a few issues with the webpages that need working through still

dae:
(in case you weren't already aware: https://github.com/ankidroid/Anki-Android-Backend/actions/runs/3580634081)

rsdroid/src/main/java/net/ankiweb/rsdroid/Backend.kt
mikehardy: Ah! That was the one lint caught that you mentioned. Got it

.github/workflows/build-all.yml
```diff
@@ -0,0 +1,134 @@
+name: Build AAR and Robo (all platforms)
+on:
+  workflow_dispatch:
```
mikehardy, line 3:
```suggestion
  workflow_dispatch:
    inputs:
      mavenPublish:
        description: "Publish artifacts to Maven Central? non-empty to publish, empty to skip publish"
        required: false
        type: string
```

.github/workflows/build-all.yml
```diff
@@ -0,0 +1,134 @@
...
+        uses: actions/upload-artifact@v2
+        with:
+          name: rsdroid-robo
+          if-no-files-found: error
+          path: rsdroid-testing/build/libs
+
+      # following steps only run on workflow dispatch
+
+      - name: Publish AAR to Maven
+        if: github.event_name == 'workflow_dispatch'
```
mikehardy, line 107:
```suggestion
        if: "${{ github.event.inputs.mavenPublish != '' && github.event_name == 'workflow_dispatch'}}" 
```

.github/workflows/build-all.yml
```diff
@@ -0,0 +1,134 @@
...
+        env:
+          ORG_GRADLE_PROJECT_SIGNING_PRIVATE_KEY: ${{ secrets.SIGNING_PRIVATE_KEY }}
+          ORG_GRADLE_PROJECT_SIGNING_PASSWORD: ${{ secrets.SIGNING_PASSWORD }}
+          SONATYPE_NEXUS_USERNAME: david-allison-1
+          SONATYPE_NEXUS_PASSWORD: ${{ secrets.SONATYPE_NEXUS_PASSWORD }}
+        run: |
+          ./gradlew rsdroid:uploadArchives -DtestBuildType=release -Dorg.gradle.daemon=false -Dorg.gradle.console=plain
+
+      - name: Publish JAR to Maven
+        if: github.event_name == 'workflow_dispatch'
```
mikehardy, line 117:
```suggestion
        if: "${{ github.event.inputs.mavenPublish != '' && github.event_name == 'workflow_dispatch'}}" 
```

.github/workflows/build-all.yml
```diff
@@ -0,0 +1,134 @@
...
+          export ANKIDROID_LINUX_CC=x86_64-unknown-linux-gnu-gcc
+          export ANKIDROID_MACOS_CC=cc
+          export RUST_DEBUG=1
+          export RUST_BACKTRACE=1
+          export RUST_LOG=trace
+          export NO_CROSS=true
+          ./gradlew rsdroid-testing:uploadArchives -Dorg.gradle.project.macCC=$ANKIDROID_MACOS_CC -DtestBuildType=debug -Dorg.gradle.daemon=false -Dorg.gradle.console=plain
+
+      - name: ℹ️Additional Release Instructions (requires human interaction)
+        if: github.event_name == 'workflow_dispatch'
```
mikehardy, line 133:
```suggestion
        if: "${{ github.event.inputs.mavenPublish != '' && github.event_name == 'workflow_dispatch'}}" 
```

rsdroid-testing/src/main/java/net/ankiweb/rsdroid/testing/RustBackendLoader.kt
```diff
@@ -38,33 +38,28 @@ object RustBackendLoader {
...
-     * 
+     *
      * @throws IllegalStateException OS is not Windows, Linux or macOS
      * @throws RuntimeException Failure when extracting library to load
      * @throws UnsatisfiedLinkError The library could not be loaded
      */
     @JvmStatic
     @Synchronized
-    fun ensureSetup(customPath: String?) {
+    fun ensureSetup() {
```
mikehardy, line 48: Ideally this is just a confirmation question to make sure I understand everything going on: the idea of customPath before was primarily because we had to specify where to get the aarch64 robo library, but now that we lipo that together from the cross-compiling x86-64 build in the action and publish it, it's effectively obsolete so we may remove it here?

*If* that's correct, only remaining question for me is, how does this intersect with `local_backend` gradle.properties entry?

mikehardy (changes requested):
Nice thanks for all the clarification.

I wasn't aware of the workflow action not being in the allowed actions list on the org because I always forget that. I added the ninja action now and wanted to run the workflows but remembered (before I ran them!) that it would trigger a publish.

I proposed an optional string for manual dispatch that will trigger publish, based on a tested/working implementation I did for cache clearing in the Anki-Android repo, what do you think? (it's a string not a boolean as an implementation detail, I found testing booleans in the workflow steps on the `if` to be fussy vs just testing a string). Now I realize github finally added manual cache deletion so maybe it's obsolete on the other workflow but it serves as an example here so that every manual run doesn't have the serious side effect of a publish

What do you think?

Also a question on custom rslib loading vs local build testing on the Anki-Android side

rslib-bridge/Cargo.toml
mikehardy: npm/yarn have only existed forever and rust had to invent a semver pinning / updating strategy that is different and almost exactly opposite for specific pins, nice https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements :laughing: 

So no semver specifiers means "go ahead and ingest non-breaking changes, exact version will be in lock", while adding a `^` means "this version in the toml and this version only". :heavy_check_mark: 

I'll play with it to see what happens, as I typically want the version in the toml itself to be bumped *if* a request for "latest version" does find a difference between version in toml and compatible version available. Some javascript / npmjs.com tools/incantations do that and some don't. I find skew between version specifier (in toml) and version used (in lock) to be a source of surprise that is not large but is non-zero leading to my preference there.

mikehardy:
Other than the windows failure due to a unix-ism below this appears to be mostly environment setup / validation / DX ergonomics type stuff

## linux

:heavy_check_mark: This passed testing on ubuntu 22

## macOS

:x: I am checking on macOS with `ALL_ARCHS=1 ./build-aar.sh` and I get:

> note: /Users/mike/work/AnkiDroid/Anki-Android-Backend/build/linker-wrapper/linker-wrapper.sh: line 4: python: command not found

Is there some way to force it to use python3? `python` no longer exists in base macOS distribution and is sort of ambiguous (changes from 2 to 3 have caused breaks all over in other projects...). 

I was able to workaround it with a prefix (shared here https://stackoverflow.com/a/74630547/9910298) so for me this worked:

`RUST_ANDROID_GRADLE_PYTHON_COMMAND=python3 ALL_ARCHS=1 ./build-aar.sh`

:x:  When I went to run `ALL_ARCHS=1 ./build-robo.sh` I got an error for missing cross-compiler:

>   error occurred: Failed to find tool. Is `x86_64-unknown-linux-gnu-gcc` installed?

On my local system, this had not been performed (ripped from build all workflow yml):

```
          brew tap SergioBenitez/osxct
          brew install x86_64-unknown-linux-gnu
          x86_64-unknown-linux-gnu-gcc -v
```

This is why we have the `./tools/doctor.sh` script I suppose ;-)

Perhaps build-robo could attempt to execute that with an error message indicating the missing thing if `x86_64-unknown-linux-gnu-gcc -v 2>&1 > /dev/null` had non-zero exit?

:x:  When I tried to run `./tools/doctor.sh` it now contains some obsolete things, it does not seem to have been altered to match this PR (yet another hairy yak...) as it talks about docker and references a deleted docker script

:x:  At that point `ALL_ARCHS=1 ./build-robo.sh` failed on missing mingw, covered here in yaml:

> brew install mingw-w64 && x86_64-w64-mingw32-gcc -v

Could probably have a similar "is it actually there?" check + failure (with these checks maybe killing the need for doctor, which would be nice since it would reduce duplication...)

:heavy_check_mark: Now `ALL_ARCHS=1 ./build-robo.sh` seemed to work (though I took a temporary DNS resolution failure to crates.io even :hankey: , seriously, what a pain testing all this stuff :sweat_smile: )

:x: gradle target `:rsdroid:lint` seems to have a silent failure here, could be peeled to a separate multi-issue (fix silent failures passing + fix failure) or could shave another yak here to add to the herd

```
> Task :rsdroid:lint
Could not load custom lint check jar file /Users/mike/.gradle/caches/transforms-3/40bf2c11c8f0fb09cac3b5552b0978fc/transformed/timber-5.0.1/jars/lint.jar
java.lang.NoClassDefFoundError: com/android/tools/lint/client/api/Vendor
```

:x:  took what looks like a flake failure, my machine was *very* heavily loaded and this popped out from gradle target `:rsdroid-instrumented:connectedDebugAndroidTest`:

```
net.ankiweb.rsdroid.BackendIntegrationTests > collectionIsVersion11AfterOpen[TestingAVD(AVD) - 12] FAILED 
        org.junit.runners.model.TestTimedOutException: test timed out after 3 seconds
        at com.google.protobuf.FieldSet.newFieldSet(FieldSet.java:99)
```

Maybe alter this to be 6 since it is arbitrary anyway? https://github.com/ankidroid/Anki-Android-Backend/blob/7405e454260aa1e31760e9847d84268f2b3a5dca/rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/BackendIntegrationTests.java#L45

Must be rare, hasn't happened before, but no need to fail it so fast just because I've got a billion different projects running tests at once

:heavy_check_mark: `./check-droid.sh` and `./check-rust.sh` appear to work fine at this point on macOS

## windows

:x: I required this in order to run things:

> Set-ExecutionPolicy -ExecutionPolicy Unrestricted -Scope Process

...as it would complain about the chained scripts as it called in to build web assets. I guess that could be in a doctor.bat file or perhaps build-aar.bat could check it somehow? Or just run the thing (it does not require special privileges to alter execution policy for that scope, and it's temporary + fairly focused so it does not seem like a big error from "increasing security attack surface area" perspective. Famous last words)

:x: At that point `build-aar.bat` was running along and failed on this ninja step (appears to be 13/24):

> FAILED: out/ts/lib/backend_proto.js out/ts/lib/backend_proto.d.ts

Error message contained the tell-tale unix-ism `rm out/ts/lib/backend_proto_static.js` instead of something I'd expect on windows like `del`

Could be this? https://github.com/ankitects/anki/blob/5e0a761b875fff4c9e4b202c08bd740c7bb37763/build/ninja_gen/src/node.rs#L299

I'm stuck on windows until that's resolved

dae:
> error occurred: Failed to find tool. Is x86_64-unknown-linux-gnu-gcc installed?

That's not a bad error as it is :-) My assumption was that developers will want to stick to single arch builds locally, with the multi-arch builds only being used for releases.

I've removed the obsolete stuff from the doctor script.

> gradle target :rsdroid:lint seems to have a silent failure here

The error doesn't seem to prevent the lint from running, as it caught that API issue, but I'll make a note.

> Set-ExecutionPolicy -ExecutionPolicy Unrestricted -Scope Process

Are you running this in PowerShell? I presume that's not required when running from cmd.exe, but I don't have an Android/Windows dev env set up to confirm.

> Error message contained the tell-tale unix-ism

The desktop build system relies on msys2 being installed (cygwin might work too). I'll update the docs.

rsdroid-testing/src/main/java/net/ankiweb/rsdroid/testing/RustBackendLoader.kt
dae: Yep. Default behaviour:

- Android builds download the .aar and extract the lib
- Robolectric builds download the .jar and extract the lib

local_backend behaviour:

- Locally-built .aar/.jar on filesystem is used instead.

Removed customPath behaviour:

- The file in the downloaded .jar was ignored, and was taken from the provided path instead. Since CI can now generate a .jar with all archs, the separate override is no longer needed. When testing locally, local_backend overrides the download, so we don't need an extra override.

rslib-bridge/Cargo.toml
dae: I'm not sure of the rationale for why they effectively reversed tilde and caret. I think the node style is a bit more logical as a caret points up, but that ship has sailed :-)

You can 'cargo install cargo-edit' and then use 'cargo upgrade' to update the .toml file in addition to the lock file, though unless the versions are tilde-pinned, it doesn't actually mean that older versions won't be allowed. If the lock file is deleted you'll end up with the latest version anyway; this mainly applies to niche cases where another crate in the same repo has pinned an exact older version for example. If one crate depends on "1.2.3" and another on "=1.2.2", a tilde pin would not allow them to compile together, whereas the default/caret style does. I'm guessing both versions could co-exist in a node project, but it's not possible to have multiple libraries with the same semver in Rust as they'd cause conflicts at link time.

mikehardy:
> My assumption was that developers will want to stick to single arch builds locally

Generally yes I think, but it is useful to validate (like now) and is a simple execution  / exit value check... can defer for some preening + extra yak shaves after

> Are you running this in PowerShell

Yes, I had not tried cmd.exe, I'll try it.

> The desktop build system relies on msys2 being installed (cygwin might work too). I'll update the docs.

Cheater! :laughing:

mikehardy:
I still have a problem running the sub-script on windows in cmd.exe, but worse - in cmd.exe I cannot run 

`Set-ExecutionPolicy -ExecutionPolicy Unrestricted -Scope Process` because it does not know about Set-ExecutionPolicy...

mikehardy:
Ooo - it silently failed this time when I went to run it again from PowerShell after running the Set-ExecutionPolicy. So there is some more silent failure going on - it was on the same `rm` line as I have not done msys2 yet (I wanted to retry it first - glad I did - silent failures are a problem). You might want to spin up windows VM and eat the dog food now and again :-). I don't have any "real" windows machine around but a VM with community Visual Code installed + choco etc comes in handy.

README.md
```diff
@@ -51,6 +51,16 @@ In Android Studio, choose the Tools>SDK menu option.
...
 synchronize gradle.
 
+### Windows: msys2
+
+Install [msys2](https://www.msys2.org/) into the default folder location.
+
+After installation completes, run msys2, and run the following command:
+
+```
+$ pacman -S git rsync
```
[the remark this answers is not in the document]
mikehardy: Okay, so after installing msys2 it appeared to build the rsdroid.dll but it did not successfully do all it's stuff, it failed (silently) in the end on a `cp` call or something. I dunno - windows looks busted to me - close to working, but busted

dae:
I have a Windows VM, but it does not currently have an Android dev environment installed.

The Windows build appears to succeed in CI, and I know the desktop build system works on Windows for me and a few other Anki contributors, so I'm guessing the issue you've hit is something specific to your machine setup rather than the build being broken for all. What is the error you are getting, and how does your output compare to what's shown in the CI run?

You could try enabling developer mode if it's not turned on - that was something that was required for Bazel, and perhaps it is still required now: https://www.howtogeek.com/292914/what-is-developer-mode-in-windows-10/

README.md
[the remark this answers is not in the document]
dae: msys2 was a requirement for building the desktop code even before this PR, as the build expects a few Unix tools like rm, git and rsync to be available. The build should be run in cmd.exe (I think PS works too), and it requires a Windows-native ninja and not the msys version that uses cygwin-style paths. ninja is a 300kb zip file, so it would make more sense to download it directly and throw it somewhere on the path than to download choco if the user is not already using choco.

dae:
Ok, I see what you mean. Sigh, setting up AS.

mikehardy:
Sorry for the pain, that's a bunch of sideways motion, but hopefully it helps the upstream stuff for you? Just to be clear + positive: this PR is amazing and I don't want this 1%-level fiddling to diminish that at all. Such a huge huge simplification here, thank you

dae:
Does it work correctly for you now?

.github/workflows/build-all.yml
```diff
@@ -65,7 +65,7 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
             target
-          key: ${{ runner.os }}-rust-release-v2-${{ hashFiles('rslib-bridge/**/Cargo.lock') }}
+          key: ${{ runner.os }}-rust-release-v2-${{ hashFiles('Cargo.lock', 'rslib-bridge/**', 'anki/rslib/**', 'anki/ftl/**', 'anki/proto/**') }}
```
mikehardy, line 68: Do you really want to hash every single file here instead of just the Cargo.lock files? That seems really expensive - it's already pretty expensive just doing the file scanning (in my experience it can add up to quite a bit of CI time). Best perhaps to focus these on just known Cargo.lock files that conclusively imply something has changed and warrants a cache upload?

README.md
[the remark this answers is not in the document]
mikehardy: go for msys2 😄

rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/BackendIntegrationTests.java
```diff
@@ -42,7 +42,7 @@ public class BackendIntegrationTests extends InstrumentedTest {
 
     /** Ensure that the database can't be locked */
     @Rule
-    public Timeout timeout = new Timeout(3, TimeUnit.SECONDS);
+    public Timeout timeout = new Timeout(10, TimeUnit.SECONDS);
```
mikehardy, line 45: Now that I am running these tests more regularly, it still sometimes failed, even at 30 so I widened it to 60
It's really just a "DB totally locked / total hang" fail-safe that should not be hit, so expanding duration until it does not
flake seems fine even out from previous 3 to 60. This works for me anyway

```suggestion
    public Timeout timeout = new Timeout(60, TimeUnit.SECONDS);
```

README.md
```diff
@@ -1,58 +1,157 @@
...
+
+or Windows:
+
+```
+set JAVA_HOME=C:\Program Files\Android\Android Studio\jre
+```
+
+Now build the .aar:
+
+./build-aar.sh
```
mikehardy, line 109:
```suggestion
```
./build-aar.sh
```
```

README.md
```diff
@@ -1,58 +1,157 @@
...
+
+If you have 'python3' on your system but not 'python', you can specify
+the name:
+
+RUST_ANDROID_GRADLE_PYTHON_COMMAND=python3 ./build-aar.sh
```
mikehardy, line 114:
```suggestion
```
RUST_ANDROID_GRADLE_PYTHON_COMMAND=python3 ./build-aar.sh
```
```

README.md
```diff
@@ -1,58 +1,157 @@
...
+
+Assuming success, then build the .jar file:
+
+./build-robo.sh
```
mikehardy, line 118:
```suggestion
```
./build-robo.sh
```
```

mikehardy (changes requested):
Tested and working on windows now! Excellent.

Made some display suggestions on README I noticed when reading the file full/separate to see it render. They are trivial

Down to just the question on cache hashing all files vs just lock files and the also-trivial extra expansion on test file timeout

This is really really close to going.

Should we just merge this and do the anki sha bump later? As long as we don't force a publish with manual workflow run + optional argument, I think it is safe to have it merged. And it clears the decks on this one

.github/workflows/build-all.yml
dae: I don't have a preference. The lockfiles should be sufficient of the majority of the cached content; changes to the other files only affect the compilation output of our first-party crates.

dae:
No objections to merging earlier, though currently the anki submodule points to a commit that's sitting in an unmerged PR on the ankidroid/anki repo. Come release time that branch will be rebased, and I'm not sure if github keep the old commit indefinitely or not. If they don't, it's possible the merged changes will fail to build in the future if the pending PR is not merged into a branch/tag.

mikehardy:
Ah! I thought it was pointing to main but just wasn't where you wanted it yet. If it's pointing to a branch better to wait for reasons you state (commit hashes might move around)

mikehardy (approved):
Worked this thing over fully, that's for sure. Looks great, just waiting on a stable anki submodule sha then. What a great change to get rid of all the Docker / cross stuff, wow. It was good while it lasted but now that M1 has fully permeated the ecosystem + going with most people just wanting to build their arch, this is so much better.

dae:
Ok, good to go. https://github.com/ankidroid/anki/pull/6 should be pulled first. Recommend a squash merge for this one, then triggering the release workflow.

mikehardy:
Fantastic! That was a pretty big changelog on the anki side (I read through almost all of it and ran out of stamina!)
Thanks again for working it through here, I'll see if I can't grab the baton and take it the rest of the way

mikehardy:
"all" workflow running now via manual dispatch with string "shipit" (:shipit: !) it takes forever so there might be a delay but I'll re-check at a later date to finalize the maven repo and everything. Hopefully it works...

mikehardy:
closed + released, maven's got the new artifact, 🏃‍♂️ is what I look like as I mosey on over to the ankidroid side to ingest the artifact and get this out. Vamos!

PR #249, opened 2022-12-20
Fix syncing with legacy_schema = false

dae:
Sorry Mike, unfortunately SSL support was missing from the 0.1.16 release, meaning users will not be able to sync with legacy_schema=false. To avoid problems for alpha testers, it might be a good idea to either:

a) do another release (and then bump the version in the AnkiDroid) before the next alpha build, or
b) revert the two commits to the AnkiDroid repo until you're ready to make another backend release

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -19,7 +19,7 @@ android.useAndroidX=true
 android.enableJetifier=false
 
 GROUP=io.github.david-allison-1
-VERSION_NAME=0.1.16-anki2.1.55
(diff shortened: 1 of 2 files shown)


mikehardy:
I haven't done a new alpha but am building to it, I'll try to roll forward by doing a 0.1.17-2.1.55 unless it fails spectacularly for some reason

mikehardy (approved):
seems simple enough :-)

mikehardy:
Works great in the app after release here / ingestion there, thanks again Damien

PR #251, opened 2022-12-23
Fix incorrect timezone on some devices; bump version

dae:
The timezone dependency Anki uses recently changed the way they gather timezone info. They used to rely on a libc call, which could cause crashes if env vars happened to be modified in another thread at the same, so they decided to reimplement the functionality in Rust instead to address a CVE: https://github.com/chronotope/chrono/issues/602

Android appeared to be supported (https://github.com/chronotope/chrono/releases/tag/v0.4.21), but it seems that this only applies to some devices, or was not tested. On a Pixel 2/API 31 sim, the /system/usr/share/zoneinfo path it refers to exists, but it does not have the individual timezone files that the library expects to find.

Until a better solution can be found, this commit reverts to an earlier version of the library where TZ lookup was working. It depends on https://github.com/ankidroid/anki/pull/7

Edit: issue reported on https://github.com/chronotope/chrono/issues/922

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,15 +81,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "android_system_properties"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
-dependencies = [
- "libc",
-]
-
-[[package]]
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
verified cargo.lock changes are the ones I'd expect from the different dep lists in the Cargo.toml, verified the commit SHA of the anki submodule is the one I fast-forward merged (to maintain the stable SHA...) and pushed out, version looks great. Good to go

PR #257, opened 2023-01-03
Switch submodule to Anki repo; bump version

dae:
As of https://github.com/ankitects/anki/commit/0eddb25287264fd9bd6beab09df691e4afb5f370, the AnkiDroid-specific patches atop the desktop repo have been merged into the desktop repo, so the anki submodule in this repo can now point directly to the main branch of the desktop code, and updating to newer desktop code just requires bumping the commit ref in the submodule here.

~No significant changes to the Rust code are expected before a 2.1.56 release, but I've added 'pre' to the version number to reflect that it's not pointing to an official release tag.~

Other changes:
- The desktop switched to a different logging library, which meant some changes here were required.
- The "fatal error" synthesis test been removed, as the PR above removed that AnkiDroid-specific error from Anki's standard error type, meaning it can't be easily synthesized from the backend method (the error is created by code in this repo, not in that repo, so mock errors need to be generated in this repo if desired in the future).

PR #268, opened 2023-01-31
Update to Anki 2.1.57

dae:
Builds upon #257 

Apart from updating to the 2.1.57 code, this PR also uses new code @Rumovz has written for the Rust timezone library:

Chrono switched away from the libc timezone implementation because of
race conditions in the env var handling, but this broke timezone handling
on Android, as Android stores timezones differently. Rumo has written
a parser for Android's custom tzdata file, which we'll contribute upstream
after testing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,10 +77,8 @@ jobs:
             anki/out/extracted
           key: ${{ runner.os }}-anki-v5-${{ hashFiles('Cargo.lock', 'anki/yarn.lock') }}
           restore-keys: |
-            Linux-anki-v5
-            Linux-anki
-            macOS-anki-v5
-            macOS-anki
+            ${{ runner.os }}-anki-v5
+            ${{ runner.os }}-anki
 
       - name: Build Android/X86_64
         if: matrix.os != 'windows-latest'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,8 +77,10 @@ jobs:
             anki/out/extracted
           key: ${{ runner.os }}-anki-v5-${{ hashFiles('Cargo.lock', 'anki/yarn.lock') }}
           restore-keys: |
-            ${{ runner.os }}-anki-v5
-            ${{ runner.os }}-anki
+            Linux-anki-v5
+            Linux-anki
+            macOS-anki-v5
+            macOS-anki
 
       - name: Build Android/X86_64
         if: matrix.os != 'windows-latest'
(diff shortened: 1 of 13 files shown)


dae:
It looks like there's a Windows-specific issue that needs to be addressed.

dae:
Appears to have been an issue with a stale cache - the scheduled builds were failing when run on the main branch as well.

dae:
Caching issue was fixed by excluding node_modules from the Windows cache. It could potentially be included in the future if the the build script was updated to recreate the junction if missing, and the cache backed up anki/node_modules instead of anki/out/node_modules.

mikehardy:
> excluding node_modules from the Windows cache.

Interesting! Usually it's nice to include that because it is a rather expensive bit of filesystem work but I'm definitely on the side of correctness before performance - thanks Damien

mikehardy (approved):
bummer about the need to duplicate the caching logic almost completely just to have effectively one line different for the windows one but I also didn't see a way in the docs to elegantly combine a matrix exclusion and the cache path definition, so it goes

Really nice simplification just pointing upstream, I suppose this means I can archive the ankidroid/anki repo since we're normally fork-based so any future patches (if needed for temporary integration) could be integrated as a hash on the fork that is the source of the PR that will go in ankitects/anki ?

mikehardy:
I would like to test this locally but I don't see any reason it won't merge at this point in case someone beats me to it and needs/wants to merge

dae:
It's not a bottleneck at the moment anyway - the Ubuntu platform finishes quicker for the feedback loop, and a full run is held up by the much slower all-platforms step.

And yeah, once AnkiDroid is updated to use the new backend release, and maybe after a new alpha's out and you're sure that you won't need to revert, then it should be safe to archive that repo. With an open PR to the main repo you can easily change the commit hash, and it's not hard to change the repo location again if necessary.

dae:
Would be great if we could get a new alpha out in the not too distant future, as alpha 92 is using GMT instead of the local timezone for calculating 'next day starts at', which is causing problems for some alpha testers, such as daily counts being reset when syncing (most recent report: https://forums.ankiweb.net/t/syncing-issue-between-ankiweb-and-mac/26374/4). This was fixed in https://github.com/ankidroid/Anki-Android-Backend/commit/9f5482570e5e9009acae70c624cc61fc00078886, but I fear I didn't adequately explain the implications of the bug at the time. To resolve the issue, either a new backend release based on this PR, or one based on the current main branch should suffice.

mikehardy:
Working through current show-stoppers + things I can immediately merge right now and will generate an alpha yep
ETA probably a day. Goal is for it to have the new backend

dae:
Thanks Mike

mikehardy (approved):
checks out locally, nice!

PR #273, opened 2023-02-27
fix(anki): bump anki commit hash to 2.1.60 tag sha

mikehardy:
I *believe* pending review that all I need to do to adopt a new upstream anki version now is switch the submodule sha to the new commit reference - if so, this is the PR that does that for anki 2.1.60

Note: this is an experiment to see if I can gain self-sufficiency in this area, I don't believe 2.1.60 is vital for AnkiDroid but the low-importance is a feature for this experiment

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Install Xcode/Visual Studio if on macOS/Windows.
 
 ### Rust
 
-Install rustup from https://rustup.rs/
+Install rustup from <https://rustup.rs/>
 
 ### Ninja
 
@@ -39,7 +39,7 @@ Windows if using choco:
 
   choco install ninja
 
-You can alternatively download a binary from https://github.com/ninja-build/ninja/releases
+You can alternatively download a binary from <https://github.com/ninja-build/ninja/releases>
 and put it on your path.
 
 ### NDK
@@ -58,7 +58,7 @@ Install [msys2](https://www.msys2.org/) into the default folder location.
 After installation completes, run msys2, and run the following command:
 
 ```
-$ pacman -S git rsync
+pacman -S git rsync
 ```
 
 When following the build steps below, make sure msys is on the path:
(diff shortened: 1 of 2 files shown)


dae (approved):
LGTM

dae:
(you'll also need to bump the backend version, but that can be done in a separate commit if you prefer)

mikehardy:
> (you'll also need to bump the backend version, but that can be done in a separate commit if you prefer)

That'd be a self-sufficiency fail this time, forgot about it. Obvious in hindsight though - thank you!
I'll bump the version in one commit and make sure this is documented somewhere now

mikehardy:
Okay - three commits now

1- the new SHA *and* the version bump (🤦)
2- auto-lint on README before I edit it
3- added a chunk containing my currently-known steps for making a formal release https://github.com/ankidroid/Anki-Android-Backend/pull/273/commits/a90c63c1cd9cf1d7d0cab55b56ad702083bfc027

item 3 there is of great interest as I clearly needed to write it down, would love if you had a moment to scan the steps and see if you remember anything that I didn't think of @dae? 🙏

mikehardy:
CI "good enough" - failure was a timeout on one side of the PR (the push) while the PR run of it actually worked. Good to go pending review

mikehardy:
I'm going to go ahead and do merge this but obviously post-comments are always welcome, easy to fix up docs

README.md
```diff
@@ -153,9 +152,17 @@ Only the current platform is built by default. In CI, the .aar and .jar files
 are built for multiple platforms, so one release library can be used on a variety
 of devices. See .github/workflows for how this is done.
 
+### Creating and Publishing a release
+
+1. Most likely you will want to align the `anki` submodule SHA with a new tagged release from upstream `ankitects/anki` repository
+1. Edit the file `gradle.properties` - increment the Anki-Android-Backend version (first part of version string) if there are code changes in this repository, and align the second part of the version string (the anki upstream part) with the tag name of the upstream tag used for the anki submodule SHA here
```
dae, line 158: You could argue that bumping the commit hash is a code change, so I usually change both versions at the same time. But as long as the total string is unique, it shouldn't block you being able to update the version.

mikehardy: I think the SHA is just like bumping a dep, it doesn't qualify as a code change here, to me. And yes because the whole string was different it worked out great

PR #299, opened 2023-08-23
fix: adopt new publishing plugin config requirements

### LINKED ISSUE #298: Maven publish gradle tasks not initialized
Something about the PRs between the last public release (0.1.21-anki2.1.61 - visible here https://repo1.maven.org/maven2/io/github/david-allison-1/anki-android-backend/) and the current state of main after merging #294 made the maven upload tasks disappear

This command is part of the release workflow:

> ./gradlew rsdroid:uploadArchives -DtestBuildType=release -Dorg.gradle.daemon=false -Dorg.gradle.console=plain

...and if you run `./gradlew -q rsdroid:tasks --all` you won't see an `uploadArchives` task anymore

Seen in this attempt to publish here: https://github.com/ankidroid/Anki-Android-Backend/actions/runs/5953436455/job/16149270782#step:13:22

It does show up just prior to merge of #294 so I think it is the changes in the last merged PR that are causing the issue

mikehardy:
the publishing plugin we use has a number of breaking changes in its current version (0.25.3) versus the previous version we used (0.14.2)

- we need to use new environment variables (version 0.15.0)
- we need to use a new target to publish (version 0.17.0)
- we need to add maven central and signing explicitly (version 0.21.0)
- we no longer need to close the repository (automatic if releasing, in 0.22.0)
- we may release the repository w/new config param (available in 0.22.0)

Fixes #298

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -89,20 +89,20 @@ jobs:
       - name: Publish AAR to Maven
         if: "${{ github.event.inputs.mavenPublish != '' && github.event_name == 'workflow_dispatch'}}"       
         env:
-          ORG_GRADLE_PROJECT_SIGNING_PRIVATE_KEY: ${{ secrets.SIGNING_PRIVATE_KEY }}
-          ORG_GRADLE_PROJECT_SIGNING_PASSWORD: ${{ secrets.SIGNING_PASSWORD }}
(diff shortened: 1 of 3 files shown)


mikehardy:
I can only test this by running it so I'll merge this assuming CI passes in general, then I'll re-try release

mikehardy:
no total failure, only valid thing is to run it so merging it to run it

PR #300, opened 2023-08-24
adopt new publishing DSL for rsdroid-testing to match rsdroid

### LINKED ISSUE #298: Maven publish gradle tasks not initialized

mikehardy:
The previous change appeared to work well with regard to the rsdroid module, however I forgot that the rsdroid-testing module has a separate build file and needs the newly-required publishing DSL as well

Hopefully this solves it too

Incidental fix to the docs, and a bump to the module version so we may run the full publish again and hopefully see it go green this time 🤞 

Fixes #298

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ export ANDROID_NDK_HOME=$HOME/Library/Android/sdk/ndk/25.2.9519653
 Or Windows using Powershell:
 
 ```
-$env:ANDROID_NDK_HOME="env:ANDROID_SDK_ROOT\ndk\25.2.9519653"
+$env:ANDROID_NDK_HOME="$env:ANDROID_SDK_ROOT\ndk\25.2.9519653"
 ```
 If you don't have Java installed, you may be able to use the version bundled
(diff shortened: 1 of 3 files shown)


mikehardy:
Only way to test it is to run it, let's go!

PR #303, opened 2023-08-24
Update all the things

mikehardy:
- update Android APIs for build/target to 34
- update gradle dependencies where possible (everything but kotlin and sqlite - should log issues for them)
- update github actions for CI
- update CI macOS runner to 13
- update ninja used in CI
- update Cargo deps for rust
- update gradle
- update android gradle plugin

All seems to work locally and in Anki-Android testing locally


mikehardy:
@dae specific request for review here is on the general idea of running `cargo update` and committing that. Seems like it should be generally okay (assuming tests pass...) and was prompted via this dependabot security alert for the repo: https://github.com/ankidroid/Anki-Android-Backend/security/dependabot/42

dae:
Yep, 'cargo update' is fine. I think you previously mentioned that you'd like to see the version numbers bumped in Cargo.toml as well; if you want to do that as well, you'd do 'cargo install cargo-edit' and then 'cargo upgrade'.

mikehardy:
> Yep, 'cargo update' is fine. I think you previously mentioned that you'd like to see the version numbers bumped in Cargo.toml as well; if you want to do that as well, you'd do 'cargo install cargo-edit' and then 'cargo upgrade'.

Oh yeah! Yes, that's a mild preference of mine so that everything ratchets up with no allowed back-sliding. It's a little obsessive but not technically wrong ;-). I'll give that a try after I merge the CI speedups and rebase on that so I don't have to wait quite as long for a fresh run here...

Thanks for the review, much appreciated.

mikehardy:
Cache behavior is *exactly* as expected / hoped-for from #304 

Cache usage now is around 8GB of the 10GB limit *with only main cached* - just no way that was going to work well when caches are not shared from branch to branch or branch to main, only from main to main or main to branch.

I believe the "release" workflows will be a lot faster now, the quick workflows will be just a little faster, and their times should be stable / consistent. That's nice so you know when to come back and check on them

mikehardy:
release workflows taking 1hr 10mins-ish, not bad at all - and this was with *lots* of recompile needed due to rust crate updates and java toolchain updates. Should improve further for future PRs once those are cached on main post-merge

PR #304, opened 2023-08-24
Improve ci caching

mikehardy:
CI is still pretty slow here on the backend repo, and I think the current two biggest items are:

- cache eviction is rapid + unwanted (seen by looking at caches for the repo - https://github.com/ankidroid/Anki-Android-Backend/actions/caches)
- nothing about java work is cached

So strategy here is to only save caches when building on main, and to add caching for gradle work. This should be a big help because caches are scoped by key+branch, so storing enormous branch caches is evicting the main cache (useful for everyone, normally) with branch caches which are not useful on main and not useful on other branches

If this works, we should see cache restores (or attempted restores) in the CI runs here for the current key with hashes plus the backup restore keys, and we should *not* see caching saves in these runs pre-merge because they are not on main.

Then post-merge we should see cache saves.

Incidentally, you can use this page to purge caches now, it's not the newest feature in github but it is relatively new: https://github.com/ankidroid/Anki-Android-Backend/actions/caches

That saves you from having to bump version-strings in cache keys normally, although that's a useful strategy as well

### DIFF
--- a/.github/workflows/build-quick.yml
+++ b/.github/workflows/build-quick.yml
@@ -72,6 +72,16 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-rust-debug-v6
 
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@v2
+        timeout-minutes: 5
+        with:
+          # Only write to the cache for builds on the 'main' branches, stops branches evicting main cache
+          # Builds on other branches will only read from main branch cache writes
+          # Comment this and the with: above out for performance testing on a branch
+          cache-read-only: ${{ github.ref != 'refs/heads/main' }}
+          gradle-home-cache-cleanup: true
+
       - name: Build all (current platform)
         run: cargo run -p build_rust
 
--- a/.github/workflows/build-quick.yml
+++ b/.github/workflows/build-quick.yml
@@ -42,8 +42,8 @@ jobs:
           distribution: "adopt"
           java-version: "17" # matches Anki-Android
 
-      - name: Rust Cache (Windows)
-        uses: actions/cache@v3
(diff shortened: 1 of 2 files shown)


dae:
IIRC node_modules was only an issue on Windows

One downside of this approach is that for backend updates (where the Rust dependencies tend to get bumped), if the PR goes through a series of changes, none of them will benefit from a cache. But hopefully we don't have too many more large PRs at this point :-)

mikehardy:
The release workflow did not restore from cache even though it appears to be looking for a restore key that exists, so, that's odd.

The windows quick run on push failed because of a network error (I've seen this before in this repo - looks like yarn has network issues semi-frequently, this is the most frequent "flake" at the moment). Not this PR's fault.

mikehardy:
> One downside of this approach is that for backend updates (where the Rust dependencies tend to get bumped), if the PR goes through a series of changes, none of them will benefit from a cache. But hopefully we don't have too many more large PRs at this point :-)

In practice, the caches here were likely evicting nearly every single time, as the 10GB limit was only sufficient to contain a single run of release/quick-from-3-OSes. Even the scheduled build from main would contend with a single branch run and evict things.

So I *believe* based on the 10GB cache limit and the size of the caches from a single run, that even though we will only have cache-from-main now, the general case and likely even the specific case you mention of rust updates etc will result in quite a bit more caching just from shared-even-if-outdated-from-specific-dependency caching from the main caches

mikehardy:
Ok then, I think I have this sorted out and the CI runs have shown me what I expect based on current understanding with regards to restore and save behavior. Going to merge this then watch those workflow runs like a hawk on main and make sure we get fresh caches there

PR #306, opened 2023-08-26
Avoid usage of Android-specific Context

dae:
This allows libanki code to be tested outside Robolectric, leading to performance improvements. Unfortunately this also requires another backend release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,16 +31,15 @@ Install rustup from <https://rustup.rs/>
 
 ### Ninja
 
-#### Debian/Ubuntu:
+#### Linux/macOS:
 
-    sudo apt install ninja-build
+Anki can be built with Ninja or N2. N2 gives better status output.
 
-#### macOS:
+  anki/tools/install-n2
 
-    brew upgrade
-    brew install ninja
+#### Windows:
 
-#### Windows if using choco:
+If using choco:
 
     choco install ninja
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -19,7 +19,7 @@ android.useAndroidX=true
 android.enableJetifier=false
 
 GROUP=io.github.david-allison-1
-VERSION_NAME=0.1.24-anki2.1.66
+VERSION_NAME=0.1.25-anki2.1.66
 
 POM_INCEPTION_YEAR=2020
 
--- a/rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/BackendDisposalTests.java
+++ b/rsdroid-instrumented/src/androidTest/java/net/ankiweb/rsdroid/BackendDisposalTests.java
@@ -42,7 +42,7 @@ public class BackendDisposalTests extends InstrumentedTest {
 
             Timber.d("Iteration %d", i);
             try (Backend backend = super.getBackend("initial_version_2_12_1.anki2")) {
-                SupportSQLiteDatabase db = AnkiSupportSQLiteDatabase.withRustBackend(BackendFactory.getBackend(getContext()));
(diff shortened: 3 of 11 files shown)


david-allison (approved):

README.md
```diff
@@ -31,21 +31,15 @@ Install rustup from <https://rustup.rs/>
...
-    brew install ninja
+#### Windows:
 
-#### Windows if using choco:
-
-    choco install ninja
-
-You can alternatively download a binary from <https://github.com/ninja-build/ninja/releases>
-and put it on your path.
+Run the command inside anki/tools/install-n2
```
mikehardy, line 42: @dae 🤔 so I did that, and now I get this when I run `.\build.bat`

```
n2: error: The filename, directory name, or volume label syntax is incorrect. (os error 123)
```

mikehardy: Looks to be a problem with string quoting related to paths, single vs double, the command that failed was

> Error: Fail with code Some(1): cmd /c tools\ninja.bat extract:protoc ts:reviewer:reviewer_extras_bundle.js qt:aqt:data:web:pages

I'm not sure of all the build script intricacies but I'm diving in. I can't recommend a toolchain change though unless it's been tested+working. You still have a windows environment around ?

mikehardy (changes requested):
Doesn't work on windows as documented, because of n2 issues, and the failure is silent.

Suggest add this patch to github workflows to expose the error:

```diff
diff --git a/.github/workflows/build-quick.yml b/.github/workflows/build-quick.yml
index 1aa8ece..e5661cc 100644
--- a/.github/workflows/build-quick.yml
+++ b/.github/workflows/build-quick.yml
@@ -36,6 +36,16 @@ jobs:
       - name: Make path absolute
         run: echo $(pwd)/ninja_bin >> $GITHUB_PATH

+      # Install n2 as recommended in documentation
+      - name: Use n2 for build (unix)
+        if: matrix.os != 'windows-latest'
+        run: ./anki/tools/install_n2
+
+      # Install n2 as recommended in documentation (note, should be in anki/tools/install_n2.bat IMHO
+      - name: Use n2 for build (windows)
+        if: matrix.os == 'windows-latest'
+        run: cargo install --git https://github.com/ankitects/n2.git --rev 89ca7dd54f6306f25eb0bfd14c7a458dd4befb7e
+
       - name: Configure JDK
         uses: actions/setup-java@v3
         with:
```

and then not sure exactly how to fix. I was able to get a little farther by running `anki\tools\ninja.bat` with a patch where I dropped the `dp0` path bit from line 3 but then hit another/similar pathing issue and ran out of time box

README.md
dae: That comes as a minor surprise, as we use n2 on Anki's CI. Don't have time to investigate at the moment, so have just rolled back the recommendation for Windows for now.

dae:
That'll probably add at least a few extra minutes to the build, unless we precompile the binaries and host them somewhere.

mikehardy:
🤔 aren't they cached along with other items? the caching is solid now...

dae:
'cargo install' is a bit different to 'cargo build'; it builds in release mode and puts the binaries in a different folder, and I think the temporary files are not stored in target/

mikehardy (approved):
works for me - as long as we're not selling n2 as the only way, we're not steering windows people into certain build-doom ;-)

mikehardy:
anyone feel free to merge once it goes green, then if you have the ability (? I think David and Dae you both do?) to run the release workflow, trigger it manually with a string as indicated, and later on it'll be up on maven...

Otherwise I'll get it tomorrow my time

dae:
Was successfully able to deploy the new library :+1:

PR #307, opened 2023-09-20
Update to Anki 23.10

dae:
Drafting until the 23.10 release


dae:
While 23.10 likely won't be stabilized until the end of the month, it would be nice to get this into the hands of users now, as FSRS works best when all clients support it. If possible, it would be great if we could publish this now, merge https://github.com/ankidroid/Anki-Android/pull/14456 in, and then put out 2.17alpha2. When 23.10 is released as stable, I'll follow-up with another backend PR that pulls in any remaining fixes accumulated since now.

david-allison:
@dae could you rebase onto `main` (now the Windows path fix is in) and I'll get this rebased + merged.

dae:
Done

david-allison (approved):

PR #318, opened 2023-10-31
test(ci): only run CI on push to main

mikehardy:
this avoids running CI on pushes to all PR branches which results in a lot of extra / unnecessary burnt CI time on PRs

Noticed by @dae in https://github.com/ankidroid/Anki-Android-Backend/pull/302#issuecomment-1692045781

This is the way we handle it in the main repo, seems to work well


dae:
LGTM

PR #320, opened 2023-11-06
Fix backend version var name in readme

abdnh:


mikehardy (approved):
Going to trust you on this one since you *just* went through this - thanks!

PR #323, opened 2023-11-17
ignore 'checkMainThreadOp' unit test error

david-allison:
Unit tests are much faster if they don't depend on Robolectric, the downside is that we can't depend on Android functionality

Looper.getMainLooper().thread is run only in `checkMainThreadOp` and this is a logging function, so we can ignore the error

```
java.lang.RuntimeException: Method getMainLooper in android.os.Looper not mocked. See https://developer.android.com/r/studio-ui/build/not-mocked for details.
		at android.os.Looper.getMainLooper(Looper.java)
```


mikehardy (approved):
🤔  can't find a hole in the logic, works for me

mikehardy:
If you want a release with this you'll need to rev the patch number in the backend-specific part of the version, then after that's committed run the github action with a workflow arg as directed in the action param text box (but no manual intervention after running the action now!) - then ingest in build.gradle on the AnkiDroid side

May be just telling you stuff you already know, but just in case it's useful...

PR #325, opened 2023-11-19
Bump version to 0.1.29-anki23.10.1

david-allison:
Would be useful to get #323 in

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -19,7 +19,7 @@ android.useAndroidX=true
 android.enableJetifier=false
 
 GROUP=io.github.david-allison-1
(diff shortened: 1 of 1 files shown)


dae (approved):

mikehardy:
@david-allison CI is quiet now with this in there. I could hit the release workflow but I'd prefer you did it just to make 100% sure I'm not the only one who has driven a release in its current incarnation - I don't think you've done so since last changes

david-allison:
@mikehardy 

I released, but the scheduled CI cancelled the run
 https://github.com/ankidroid/Anki-Android-Backend/actions/runs/6923692327

PR #326, opened 2023-11-20
Workflow actions update (lint, bump versions, use n2, public release new concurrency group)

mikehardy:
This was mainly prompted by https://github.com/ankidroid/Anki-Android-Backend/pull/325#issuecomment-1818743536

> I released, but the scheduled CI cancelled the run

Cleaned up a few other things while there

Related https://github.com/ankitects/anki/pull/2837 for a windows n2 install batch script

Learnings:
"concurrency" in github actions is just a string. So whatever string you put in *is* the key used for concurrency lookup when applying concurrency rules if I understand correctly. Thus manipulating it as proposed here should work fine 

https://docs.github.com/en/actions/using-jobs/using-concurrency

> The concurrency key is used to group workflows or jobs together into a concurrency group. When you define a concurrency key, GitHub Actions ensures that only one workflow or job with that key runs at any given time

### DIFF
--- a/.github/workflows/build-quick.yml
+++ b/.github/workflows/build-quick.yml
@@ -5,7 +5,7 @@ on:
     - cron: "0 0 * * *"
   pull_request:
     branches:
-      - '**'
+      - "**"
   push:
     branches:
       - main
--- a/.github/workflows/build-quick.yml
+++ b/.github/workflows/build-quick.yml
@@ -31,23 +31,6 @@ jobs:
       - name: Fetch submodules
         run: git submodule update --init
 
-      - name: Setup Ninja
-        uses: ashutoshvarma/setup-ninja@master
-        with:
-          version: 1.11.1
-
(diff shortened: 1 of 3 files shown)


mikehardy:
Looking for a review from David on the concurrency thing
Looking for a review from Damien on the switch to n2 in CI as an idea

david-allison (approved):
Concurrency & checkout LGTM

mikehardy:
given n2 worked for windows (hooray!) I still would like a PR-post-merge check from Damien if possible but I think this is good to go. Merging so we can do the release, and it will inadvertently test the concurrency as the release build all and merge build all should go at same time

mikehardy:
new concurrency worked fine, my manual release did not concurrency-kill the merge ("push" trigger) run of the workflow ✅

dae:
LGTM; have left a comment [here](https://github.com/ankitects/anki/pull/2837#issuecomment-1820156221)

PR #327, opened 2023-11-21
build: use bash for install-n2 on all platforms

mikehardy:
semi-speculative test. will `bash` (vs `bash.exe` or `gitbash.exe` etc) work on windows, allowing the same run command to work for all platforms on the install-n2 step?


mikehardy:
Works great - a unix and a windows runner have both made it past the new unified install-n2 step advised by Damien on #326 - so this cleans up the workflow and removes any knowledge of commit hashes here, all defers to upstream as it should

PR #331, opened 2023-12-01
build(deps): adopt sqlite 2.4.0

mikehardy:
they converted to Kotlin upstream, so some methods became properties and some parameters had nullability stuff change

this passes testing in-repo and with `local_backend=true` in peer-level Anki-Android repo


david-allison (approved):
LGTM

we're heavily tested on the Android side which will confirm functionality is still compatible

Should have been a 3.0 release given the breaking changes

PR #335, opened 2023-12-12
Ship root-vars, reviewer.css and reviewer.js

BrayanDSO:
Necessary for the Previewer/Reviewer port.

Pardon if I did any rust mistakes

Unblocks https://github.com/ankidroid/Anki-Android/pull/14949

### DIFF
--- a/build_rust/src/main.rs
+++ b/build_rust/src/main.rs
@@ -61,6 +61,9 @@ fn build_web_artifacts() -> Result<()> {
     cmd.current_dir("anki")
         .args([
             "extract:protoc",
+            "css:_root-vars",
+            "ts:reviewer:reviewer.js",
+            "ts:reviewer:reviewer.css",
             "ts:reviewer:reviewer_extras_bundle.js",
             "ts:reviewer:reviewer_extras.css",
             "qt:aqt:data:web:pages",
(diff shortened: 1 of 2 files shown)


build_rust/src/main.rs
```diff
@@ -78,14 +78,18 @@ fn build_web_artifacts() -> Result<()> {
...
         )?;
     }
+    for file in read_dir_files("anki/out/ts/reviewer")? {
+        let file = file?;
+        let path = file.path();
+        copy_file(
+            &path,
+            web_dir.join(path.file_name().unwrap().to_str().unwrap()),
+        )?;
+    }
```
dae, line 88: I have a minor preference for manually listing out the files, instead of iterating. That avoids including unwanted files like .map in debug builds, and makes it explicit which files are being required (and throwing an error if they happen to be missing).

dae: Also, reviewer_extras contains a subset of reviewer.css. You probably don't need both.

build_rust/src/main.rs
```diff
@@ -78,14 +78,18 @@ fn build_web_artifacts() -> Result<()> {
...
-        "anki/out/ts/reviewer/reviewer_extras_bundle.js",
-        artifacts_dir.join("web/reviewer_extras_bundle.js"),
-    )?;
-    copy_file(
-        "anki/out/ts/reviewer/reviewer_extras.css",
-        artifacts_dir.join("web/reviewer_extras.css"),
-    )?;
+            "anki/out/qt/_aqt/data/web/css/webview.css",
+            artifacts_dir.join("web/webview.css"),
+        )?;
```
dae, line 92: css:_root-vars.css and out/sass/_root-vars.css may be a better choice

build_rust/src/main.rs
BrayanDSO: The same applies for reviewer.js and reviewer_extras_bundle.js?

dae: Ah, but if you're planning to have the old and new review screen be toggleable with a setting, then I guess you'll need to include both versions for now, and choose the one to use based on that setting.

mikehardy (approved):
seems okay - but as long as you're around @dae - you actually know the contents of these files, I don't! Feel free to do more than comment and give it a formal approve if you actually like it

@BrayanDSO note that you should probably add a second / separate commit that bumps the backend version to 0.1.32 so we're ready to release this for you and adopt it in Anki-Android repo once it's good to go

BrayanDSO:
Done

dae (approved):

dae:
I kicked off a publish; should become available a while after https://github.com/ankidroid/Anki-Android-Backend/actions/runs/7188672975/job/19578767364 completes.

PR #339, opened 2023-12-27
Update to 23.12

dae:
Will require https://github.com/ankidroid/Anki-Android/issues/15083#issuecomment-1869911711 on the frontend

Drafting because I'll need help figuring out a gradle issue. import-anki-package.html should be included in the bundle, but it's not for some reason. If you extract classes.jar from rsdroid/build/outputs/aar/rsdroid-release.aar after running build.sh, then look at the contents of classes.jar, you'll notice there's no import-anki-package.html. But if you look in rsdroid/build/generated/anki_artifacts/web/, you'll see the file is there. So the Rust build script puts it there, but for some reason gradle is ignoring it when generating the sourceSet. Any ideas? The file will be needed for properly importing .apkg files, as users need to answer questions at import time.

### DIFF
--- a/build_rust/src/main.rs
+++ b/build_rust/src/main.rs
@@ -49,7 +49,7 @@ fn run_gradle() -> Result<()> {
 
 fn build_web_artifacts() -> Result<()> {
     println!("*** Building desktop web components");
-    let artifacts_dir = Path::new("rsdroid/build/generated/anki_artifacts");
+    let artifacts_dir = Path::new("rsdroid/build/generated/anki_artifacts/backend");
     let mut cmd = if cfg!(windows) {
         let mut cmd = Command::new("cmd");
         cmd.args(["/c", "tools\\ninja.bat"]);
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1418,7 +1418,7 @@ dependencies = [
 [[package]]
 name = "fsrs"
 version = "0.1.0"
-source = "git+https://github.com/open-spaced-repetition/fsrs-rs.git?rev=f45f46bdba6625f03677eaeb039dd8a6ffcad688#f45f46bdba6625f03677eaeb039dd8a6ffcad688"
+source = "git+https://github.com/open-spaced-repetition/fsrs-rs.git?rev=58ca25ed2bc4bb1dc376208bbcaed7f5a501b941#58ca25ed2bc4bb1dc376208bbcaed7f5a501b941"
 dependencies = [
  "burn",
  "itertools 0.12.0",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,14 +18,26 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
-name = "ahash"
+name = "aes"
 version = "0.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2c99f64d1e06488f620f932677e24bc6e2897582980441ae90a671415bd7ec2f"
+checksum = "ac1f845298e95f983ff1944b728ae08b8cebab80d684f0a832ed0fc74dfa27e2"
+dependencies = [
+ "cfg-if",
+ "cipher",
+ "cpufeatures",
+]
+
+[[package]]
+name = "ahash"
+version = "0.8.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "91429305e9f0a25f6205c5b8e0d2db09e0708a7a6df0f42212bb56c32c8ac97a"
 dependencies = [
  "cfg-if",
  "once_cell",
  "version_check",
+ "zerocopy",
 ]
 
 [[package]]
@@ -124,7 +136,7 @@ dependencies = [
  "hyper",
  "id_tree",
  "inflections",
- "itertools",
+ "itertools 0.12.0",
  "lazy_static",
  "nom",
  "num_cpus",
@@ -150,7 +162,7 @@ dependencies = [
  "sha1",
  "snafu",
  "strum",
- "syn 2.0.38",
+ "syn 2.0.43",
  "tempfile",
  "tokio",
  "tokio-util",
@@ -162,9 +174,9 @@ dependencies = [
  "unicase",
  "unicode-normalization",
  "utime",
- "windows 0.51.1",
+ "windows 0.52.0",
  "zip",
- "zstd",
+ "zstd 0.13.0",
 ]
 
 [[package]]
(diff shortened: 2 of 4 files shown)


mikehardy:
Is resource shrinking enabled? This may be off target but it reminds of a frightening realization I had a long time ago and led to me actually tweeting (something I basically never do) https://twitter.com/fastsquatch/status/1196475279688028160 - perhaps we need a keep.xml to maintain resources we want?

Lots of assumptions to verify before thinking that's the thing causing the file to not make it to final artifact but maybe ?

dae:
Minify is disabled and shrink isn't mentioned, so unfortunately I don't think that's it.

mikehardy:
well, bummer. that one's an easy fix if it *is* it. I'm on a beach vacation right now and will meet my untimely demise at hands of family if I dig into this now so it'll have to sit until the new year, sorry

dae:
No worries, enjoy the holiday :-)

mikehardy:
Had some time to look into this today. Could not find a way to directly alter the gradle resource merging process to achieve the desired result (that being, `import-anki-package.html` actually showing up in the merged classes.jar...) but was able to see the problem even on `main` and note that:

1- rust build is definitely generating it, verifiable with inspection of `rsdroid/build/generated/anki_artifacts`
2- gradle build initially "sees" it, verifiable with inspaction of `rsdroid/build/intermediates/java_res/<flavor>/out/web`
3- gradle build keeps the import-anki-package js and css but drops the html as it begins merge, with `import-anki-package.html` first (I think?) missing in `rsdroid/build/intermediates/merged_java_res/<flavor>/feature-rsdroid.jar`

I think specification of resource directories is simple in the sourceSet declaration, and appears to be working well or everything about the pages would be missing, and I believe item 1 and 2 show that working - gradle initially sees the page just fine and does one copy of it into the intermediate java_res area

What mechanism defines + controls the resources merger and results in the file going missing as the merger happens? Unknown.

This is the only HTML page with two hyphens is the name, but that's ridiculous. File contents are pretty regular and the HTML looks the same as the other pages that work so I don't think it's some HTML validation failure.

Next steps will be to toggle on whatever verbosity is possible to see what/how/why gradle merges the resources, hoping for insight.

Out of time for a while tho

david-allison:
Test script

```sh
unzip -o /Users/davidallison/StudioProjects/Anki-Android-Backend/rsdroid/build/outputs/aar/rsdroid-release.aar
unzip -o classes.jar

if [ ! -f web/card-info.css ]
then
   echo "ERROR"
   exit 1
fi

if [ -f web/import-anki-package.html ]
then
   echo "import-anki-package.html exists"
else
   echo "import-anki-package.html does not exist"
fi

date -r /Users/davidallison/StudioProjects/Anki-Android-Backend/rsdroid/build/outputs/aar/rsdroid-release.aar | awk '{print substr($0,12,8)}'
```

david-allison:
```patch
Index: rsdroid/build.gradle
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/rsdroid/build.gradle b/rsdroid/build.gradle
--- a/rsdroid/build.gradle	(revision 3c4237da95a2399aa02dc7112c8460883cb8f7d2)
+++ b/rsdroid/build.gradle	(date 1704195534545)
@@ -77,7 +77,7 @@
             kotlin.srcDirs += "build/generated/source/backend"
             java.srcDirs += "build/generated/source/backend"
             jniLibs.srcDirs 'build/generated/jniLibs'
-            resources {
+            assets {
                 srcDirs "src/main/resources", "build/generated/anki_artifacts"
             }
         }
@@ -116,7 +116,6 @@
     testImplementation 'androidx.test:core:1.5.0'
     testImplementation "androidx.test.ext:junit:${rootProject.ext.androidxTestJunitVersion}"
     testImplementation project(path: ':rsdroid-testing')
-
 }
 
 tasks.withType(KotlinCompile).configureEach {
```

files are now available in `rsdroid-release.aar`: `assets/web`

mikehardy:
didn't realize resources had any sort of auto-munging going on, if that's the case and the universe of possible things headed into it (from anki upstream) is unbounded, then I like the idea of doing it through the assets path (which is, I hope, not munged in any way?). The loading seems basically the same on the app side

mikehardy:
might be best to nest it in a namespace that is clearly named for it's source, as they are built they are built in to anki_backend and that's unambiguous, easy to understand, easy to avoid for others whereas "web" is as you mentioned something that could clash.

Perhaps `assets/anki_backend` ?

BrayanDSO:
> Perhaps assets/anki_backend ?

👍

dae:
Thanks for looking into this, David! Unfortunately it doesn't seem to work for me. Any ideas what I might be missing?

```
% ./gradlew clean; touch rslib-bridge/build.rs
% ./build.sh
% unzip -l rsdroid/build/outputs/aar/rsdroid-release.aar
Archive:  rsdroid/build/outputs/aar/rsdroid-release.aar
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  1980-02-01 00:00   R.txt
      201  1980-02-01 00:00   AndroidManifest.xml
  3686180  1980-02-01 00:00   classes.jar
        0  1980-02-01 00:00   proguard.txt
      156  1980-02-01 00:00   META-INF/com/android/build/gradle/aar-metadata.properties
        0  1980-02-01 00:00   jni/
        0  1980-02-01 00:00   jni/x86_64/
 57159104  1980-02-01 00:00   jni/x86_64/librsdroid.so
---------                     -------
 60845641                     8 files
```

This is after applying the first patch in https://github.com/ankidroid/Anki-Android-Backend/pull/339#issuecomment-1873922668

dae:
Ah:

```diff
diff --git a/rsdroid/build.gradle b/rsdroid/build.gradle
index 5cc5e37c6..2a2c8ded2 100644
--- a/rsdroid/build.gradle
+++ b/rsdroid/build.gradle
@@ -81,7 +81,7 @@ android {
                 srcDirs "src/main/resources"
             }
             assets {
-                "build/generated/anki_artifacts"
+                srcDirs "build/generated/anki_artifacts"
             }
         }
     }
```

dae:
Please squash when merging.

david-allison (approved):

build_rust/src/main.rs
```diff
@@ -49,7 +49,7 @@ fn run_gradle() -> Result<()> {
 
 fn build_web_artifacts() -> Result<()> {
     println!("*** Building desktop web components");
-    let artifacts_dir = Path::new("rsdroid/build/generated/anki_artifacts");
+    let artifacts_dir = Path::new("rsdroid/build/generated/anki_artifacts/backend");
```
mikehardy, line 52: am I correct in thinking the paired PR on front-end will now require an adjustment to load everything from assets in this new sub-directory? I *think* that's pretty obvious but I don't see the changes on the related PR yet so just want to double-check

mikehardy (approved):
LGTM over here in backend-land, just a clarifying comment to verify understanding of needed changes on frontend

PR #367, opened 2024-04-09
Remap POST requests url

BrayanDSO:
Summary of the issue: 

Loading files with `file:/` scheme in much faster in an Android Webview, and that's what 2.15.6 did by setting the Webview base url to the media directory. 

But in 2.16/2.17, in order to enable POST requests to be done, the base url was set to localhost and the files were delivered by the server. After noticing that loading files was slower with the server, a `WebViewAssetLoader` was used, but it was still slower than `file:/`. The next alternative was parsing the card's HTML to use `file:/` for media sources, but that leads to CORS issues and sometimes HTML parsing issues.

So, here's what I believe that should be the definitive solution, which is to remap POST requests base url, so the Webview base url can be set to the media directory so the 2.15.6 behavior is restored.

---

Basically, this replaces `"/_anki/"` in `reviewer.js` and `reviewer_extras_bundle.js` so a custom post base url can be used if set in AnkiDroid.

My experience with Rust is almost zero, so I expect that there should be better practices than what I did.

Tested with [this PR](https://github.com/ankidroid/Anki-Android/pull/16140) by adding `console.log("foo");` into `Custom scheduling` deck option.

---

Ideally, this should be unit tested, but **I currently don't have time for that, nor for dealing with requested changes in this PR, so please take over this if time is of the essence.**

### DIFF
--- a/build_rust/src/main.rs
+++ b/build_rust/src/main.rs
@@ -2,10 +2,12 @@ use anki_io::{copy_file, create_dir_all, read_dir_files};
 use anki_process::CommandExt;
 use anyhow::Result;
 use camino::{Utf8Path, Utf8PathBuf};
-use std::env;
 use std::env::consts::OS;
-use std::path::Path;
+use std::fs::File;
+use std::io::{BufRead, BufReader, BufWriter, Write};
+use std::path::{Path, PathBuf};
 use std::process::Command;
+use std::{env, io};
 
 const ANDROID_OUT_DIR: &str = "rsdroid/build/generated/jniLibs";
 const ROBOLECTRIC_OUT_DIR: &str = "rsdroid-testing/build/generated/jniLibs";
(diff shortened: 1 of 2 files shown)


dae:
Does the rename work on Windows while the file is still open?

Doing it at build time is a valid approach, but AnkiMobile takes the simpler approach of doing the replacement when the webpage is being served to the frontend.

dae:
Windows can throw errors when you try to do something like delete a file that is currently open for writing. But if the build scripts are working on a Windows machine for you (and not WSL), then it sounds like it's fine.

build_rust/src/main.rs
```diff
@@ -109,6 +111,16 @@ fn build_web_artifacts() -> Result<()> {
         "anki/ts/licenses.json",
         artifacts_dir.join("web/licenses-ts.json"),
     )?;
+    let _ = replace_string_in_file(
```
david-allison, line 114: nit: could you provide a comment with a sample output of the replacement

david-allison (approved):
Good to go. One nice to have

mikehardy (approved):
Seems functional enough - read through the other comments. Release build is on mac so that won't be a problem in infrastructure, and if it works for the people that say they were working on it, works for me

I'm most interested in making sure that the idea actually works, and if the replacement needs to move later I think that's minor compared to getting the idea of the remap into backend/frontend

mikehardy:
I'm going to do the release process as well - that changed so recently and has only been exercised once so I'm keen to see it exercised again, especially since the version bump is already in here

PR #370, opened 2024-04-21
build(deps): update to Anki 24.04.1

david-allison:
```
cd anki
git checkout tags/24.04.1
```

https://github.com/ankitects/anki/commit/ccd9ca1a8309b80bcb50ddc5d99c7ce63440bce9

Changelog: https://github.com/ankitects/anki/releases/tag/24.04.1

Note: the git tag is a detached HEAD

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -19,7 +19,7 @@ android.useAndroidX=true
 android.enableJetifier=false
 
 GROUP=io.github.david-allison
-VERSION_NAME=0.1.37-anki24.04
+VERSION_NAME=0.1.38-anki24.04.1
 
 POM_INCEPTION_YEAR=2020
 


mikehardy (approved):
I heard rumblings that there were more things required, is this not the case? This works when used with local_backend in local.properties on the Anki-Android side? Just confirming

Seems fine otherwise

Note that version-wise, if it is literally nothing but the anki changes you don't need to bump the anki-android-backend part of the version but at the same time that doesn't matter a bit, so, all good as far as I am concerned

david-allison:
SvelteKit changes were in .2, which is currently in beta: 

* https://github.com/ankidroid/Anki-Android/issues/16232

Version: I prefer the bump on both sides, but it's stylistic, make a rule and stick with it and I'm happy

BrayanDSO (approved):

mikehardy:
bump all the versions since I don't actually care and you have a preference :)

Issue #373, opened 2024-05-03
CI Failure (ubuntu)


david-allison:
`Error: Failed with code Some(1): ./ninja extract:protoc css:_root-vars ts:reviewer:reviewer.js ts:reviewer:reviewer.css ts:reviewer:reviewer_extras_bundle.js ts:reviewer:reviewer_extras.css qt:aqt:data:web:pages`

* Last success: 2024-04-29T01:46:25+01:00 - https://github.com/ankidroid/Anki-Android-Backend/actions/runs/8871673583/job/24354979002
* First failure: 2024-04-30T01:44:57+01:00 https://github.com/ankidroid/Anki-Android-Backend/actions/runs/8887480874/job/24402827438

david-allison:
**good**
```
2024-04-29T00:49:16.7238937Z ##[group]Run cargo run -p build_rust
2024-04-29T00:49:16.7239368Z [36;1mcargo run -p build_rust[0m
2024-04-29T00:49:16.7289573Z shell: /usr/bin/bash -e {0}
2024-04-29T00:49:16.7289911Z env:
2024-04-29T00:49:16.7290151Z   DEBUG: 1
2024-04-29T00:49:16.7290539Z   ANDROID_NDK_HOME: /usr/local/lib/android/sdk/ndk/26.3.11579264
2024-04-29T00:49:16.7291302Z   ANDROID_NDK_ROOT: /usr/local/lib/android/sdk/ndk/26.3.11579264
2024-04-29T00:49:16.7291905Z   JAVA_HOME: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/21.0.2-13/x64
2024-04-29T00:49:16.7292585Z   JAVA_HOME_21_X64: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/21.0.2-13/x64
2024-04-29T00:49:16.7293147Z   GRADLE_ACTION_ID: gradle/actions/setup-gradle
2024-04-29T00:49:16.7293586Z   GRADLE_BUILD_ACTION_SETUP_COMPLETED: true
2024-04-29T00:49:16.7293990Z   GRADLE_BUILD_ACTION_CACHE_RESTORED: true
2024-04-29T00:49:16.7294584Z   DEVELOCITY_INJECTION_INIT_SCRIPT_NAME: gradle-actions.inject-develocity.init.gradle
[...]

PR #376, opened 2024-05-14
fix(ci): codeql custom build workflow needs build retry workaround as well

mikehardy:
See #373  - codeql also runs the rust build on ubuntu platform and thus needs the "retry a few times" build workaround


mikehardy:
I verified this passed the part that was still causing problems, merging so I can rebase other stuff on it

PR #387, opened 2024-06-11
feat: optionally log calls on main thread

david-allison:
Also changes default to 'off'

Previously it was required

* The logging is a slow operation
* translate() does not require it
* `browserRowForId()` probably does not require it
* This spams logs in production

We've made great progress tackling `getColUnsafe`, and just by the name, we will eventually get to 0
usages

The logging is less relevant now, especially as
AnkiDroid: 043ded78761fea83a0614206eb987eb891799b9d means that it's compiler warning
to access the collection with the main thread

Fixes 385


mikehardy (approved):
okay

Issue #399, opened 2024-08-03
Scheduled build is broken


david-allison:
Last pass:

* https://github.com/ankidroid/Anki-Android-Backend/actions/runs/10154203453

First fail:

* https://github.com/ankidroid/Anki-Android-Backend/actions/runs/10190016386

david-allison:
Likely the runner image update:

 * ✅  `Version: 20240721.1.0`
 * ❌  `Version: 20240730.2.0`

NDK update: https://github.com/actions/runner-images/commit/2a4bc14da46f1f8e358aa902a69edb9bef135472

mikehardy:
Irritating that the runner versions skew, I remember this was a problem before on some part of the toolchain may have in fact been the NDK when Damien was in here berserking all the code a few months back, but it self-resolved and I suppose we forgot about it.

Correct resolution is to manually set the version which is handled in #402, so this shouldn't recur at least

PR #402, opened 2024-08-10
Update NDK to v27, Anki to 24.06.3

voczi:
So there's also this thing about 16 kb page sizes (https://developer.android.com/guide/practices/page-sizes). I have confirmed that I get librsdroid with a 16kb page size, so that should now be fine on the Anki-Android-Backend side:
image
But I haven't really managed to get an emulator running with a 16kb page size. The one provided by Google seems to still be reporting a 4kb page size for the system, same as the regular Android 15 image? Basically, right now we can't really confirm that it runs as it should in Anki-Android under Android 15 with a 16kb page size (because no such image actually exists). But it does run as it should otherwise.

Regarding the NDK download scripts... These, along with the workflows for building a release, were refactored because the macOS image has an out-of-date NDK for now (26 instead of latest 27). So I basically made sure that in the future, in case Windows/Ubuntu gets an outdated NDK, we can just download the one version we want anyway. I would like to find a way to refactor the NDK versions in the codebase so it's no longer hardcoded. But maybe we'll leave that for another time?

Should fix:
* Fixes https://github.com/ankidroid/Anki-Android-Backend/issues/399
* Fixes https://github.com/ankidroid/Anki-Android-Backend/issues/398

### DIFF
--- a/.github/workflows/build-quick.yml
+++ b/.github/workflows/build-quick.yml
@@ -12,6 +12,7 @@ on:
 
 env:
   DEBUG: 1
+  ANDROID_NDK_VERSION: "27.0.12077973"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -50,15 +51,21 @@ jobs:
       - name: Fetch submodules
         run: git submodule update --init
 
-      - name: Set NDK version (Unix)
-        if: contains(matrix.os, 'windows') == false
+      - name: Install/Set NDK version (Unix)
+        if: contains(matrix.os, 'ubuntu') || contains(matrix.os, 'macos')
         run: |
+          export PATH="$PATH:$ANDROID_HOME/cmdline-tools/latest/bin"
+          ./.github/scripts/install_ndk.sh ${ANDROID_NDK_VERSION}
+          export ANDROID_NDK_LATEST_HOME="${ANDROID_SDK_ROOT}/ndk/${ANDROID_NDK_VERSION}"
           echo "ANDROID_NDK_HOME=$ANDROID_NDK_LATEST_HOME" >> $GITHUB_ENV
           echo "ANDROID_NDK_ROOT=$ANDROID_NDK_LATEST_HOME" >> $GITHUB_ENV
 
-      - name: Set NDK version (Windows)
+      - name: Install/Set NDK version (Windows)
         if: contains(matrix.os, 'windows')
         run: |
+          $env:PATH = "$env:PATH;$env:ANDROID_HOME\cmdline-tools\latest\bin"
+          ./.github/scripts/install_ndk.bat $env:ANDROID_NDK_VERSION
+          $env:ANDROID_NDK_LATEST_HOME = "$env:ANDROID_SDK_ROOT\ndk\$env:ANDROID_NDK_VERSION"
           Add-Content -Path $env:GITHUB_ENV -Value ANDROID_NDK_HOME=$env:ANDROID_NDK_LATEST_HOME
           Add-Content -Path $env:GITHUB_ENV -Value ANDROID_NDK_ROOT=$env:ANDROID_NDK_LATEST_HOME
 
--- a/.github/scripts/install_ndk.bat
+++ b/.github/scripts/install_ndk.bat
@@ -1,0 +1,12 @@
+@echo off
+if [%1]==[] goto usage
+
+echo Installing NDK %1
+echo %PATH%
+sdkmanager --list_installed | findstr ndk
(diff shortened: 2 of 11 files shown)


voczi:
So macOS seems to be having problems with the linker argument.. 🤔 
image

voczi:
macOS builds looking good now, too.
image

voczi:
Should fix:
https://github.com/ankidroid/Anki-Android-Backend/issues/399
https://github.com/ankidroid/Anki-Android-Backend/issues/398

voczi:
Never mind, seems like you have to install the "VanillaIceCream 16kb" image instead of the "API 35 16kb" one.
image
Seems to be running fine in Android 15 16kb. Zipalign confirms a 16kb alignment on the Anki-Android release APKs, too.
image

david-allison (commented):
Thanks!!!!

Code looks great.

A concern here is the bump to

* https://github.com/ankitects/anki/commit/872ecf481ab33b179726284b754d46ca95deda67

Rather than a tagged 'stable' release such as:

* Anki 24.06.3: https://github.com/ankitects/anki/commit/d678e39350a2d243242a69f4e22f5192b04398f2

If this is necessary, could the reason be documented in the commit message/PR

If it's not, could we only push forward to 24.06.3

I wouldn't want to release 'alpha-quality' Anki Desktop code in our stable releases.

(not that Anki Desktop Alpha quality code is anywhere near bad, but I'd rather pin to releases which upstream feels are of production quality, given they have a long/solid pre-release process).

voczi:
> If it's not, could we only push forward to 24.06.3
Works for me. Well, I can just put the first commit in a draft PR so the work done for updating the Rust stuff doesn't go to waste. I mean, the NDK stuff took me the most time. But still, could be nice to have something we can insta-merge to get back on track when Anki has another release.

david-allison (approved):
Thanks so much!

voczi:
There may have been some code duplication in the workflow... 👀 
Fixed now.

user1823:
@mikehardy, if you have the time, would you please have a look at this? I have been waiting for a release with Anki 24.06.3 since long.

.github/workflows/build-quick.yml
```diff
@@ -12,6 +12,7 @@ on:
```
mikehardy, line 12: CI should test 16kb and non-16kb page alignment in libraries, packaging, and emulator

with upcoming need to be ["16kb alignment clean"](https://developer.android.com/guide/practices/page-sizes) the CI matrix for emulator tests should be updated to run against a 16kb emulator as well as a regular emulator. Suggest using this API/image-name tuple:

android-35;google_apis_ps16k

The page size should be verified in CI prior to execution, via the tests on linked google doc

Additionally, the .apk should be extracted and have .so file verification of page-alignment checked, and finally the zipalign should be checked - otherwise we won't know if we have actually done the work or not, and we won't track any random regressions

I will create an issue from this comment (in the Anki-Android repo, as I believe it should be checked there so we cover *all* libraries not just Backend) then resolve this comment so we don't slow down this PR at all

mikehardy (approved):
Left some comments that will all be pulled to new issues, no action needed here - I want to get the PR in without delay as it functions well

Tested:
- locally after `gh pr checkout 402 --recurse-submodules`
- ./check-rust.sh worked
- ./check-droid.sh worked (with API34 regular emulator and API35_16ps emulator)
- used local backend in Anki-Android via local.properties and lib.versions.toml change

Everything checks out ✅ 

Awesome - thanks for handling not just the anki update but the 16kb page size stuff, moves us forward before it's a problem, a nice feeling

I'll publish this post merge (and after updating dependencies here) then make a PR in Anki-Android repo to reference this work

mikehardy:
(actually there is one tiny problem with `./gradlew jacocoTestReport` in Anki-Android but it looks suspiciously like a timestamp value and it looks suspiciously like exactly the sort of thing that would change as anki changes - that is, we were asserting some thing we thought was fixed but either behavior changed or the value did not mean what we thought it meant - either way, not this PRs fault I think...will pursue elsewhere

```
SchedulerTest > test_preview FAILED
    java.lang.AssertionError: expected:<1724180149594> but was:<1724180149595>
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:647)
        at org.junit.Assert.assertEquals(Assert.java:633)
        at com.ichi2.libanki.SchedulerTest.test_preview(SchedulerTest.kt:881)
```
)

PR #409, opened 2024-09-08
Update Rust to 1.81 following CVE

voczi:
See: https://vuldb.com/?id.276509
Backend affected by this because of `std::process::Command` usage, and a rustc version <1.81.0.
The `time` cargo dependency had to be updated following this change, otherwise you would get this error:
```
error[E0282]: type annotations needed for `Box<_>`
...
= note: this is an inference error on crate `time` caused by an API change in Rust 1.80.0; update `time` to version `>=0.3.35` by calling `cargo update`
```

### DIFF
--- a/.github/scripts/purge_ndk.sh
+++ b/.github/scripts/purge_ndk.sh
@@ -1,0 +1,18 @@
+: ${1?"Usage: $0 NDK Version (ex 21.3.6528147)"}
+
+echo "purging unused NDKs, keeping NDK $1..."
+
+# (i know, i know)
+brew install grep
+
+list=$(sdkmanager --list_installed | ggrep -Eo "ndk;[0-9.]+")
+for ver in $list
+do
+    if [ ! "ndk;$1" == "$ver" ];
+    then
+        echo "uninstalling ndk version $ver"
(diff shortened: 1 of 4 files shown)


Arthur-Milchior:
Mac Os failed but I could not find any cause of failure. It ended with "setup gradle" but no error message.

I restarted it just in case it's transient.
I wanted to review, but I admit I don't even know how to review this. This seems to make sens and clearly the test builds. But I've no idea how the particular hash and version number were obtained for example,

voczi:
> I wanted to review, but I admit I don't even know how to review this.

Forgot to say, Mike said he would look at it later today (if someone else could fix it in the meantime, it would be welcome I'm sure.)
But yeah, here is the error:
Screenshot_20240908_150431_Brave

dae:
Try bump the restore key: https://github.com/ankidroid/Anki-Android-Backend/blob/main/.github/workflows/build-release.yml

mikehardy (approved):
I was disappointed to see there was no generic "reclaim disk space" action like we have for ubuntu (needed in main repo) but alas there isn't

This NDK deleter is as good as any though, and nicely done how specific it is, works for me

Thank you!

PR #414, opened 2024-09-27
feat(anki): bump to anki 24.10beta1

mikehardy:
Please note this is a speculative commit.

The beta changelog upstream has a couple warnings indicating we do not want to adopt this yet.

Zero local testing has been performed at all

This is literally just bumping the submodule (with `--recurse-submodules`) and pushing it to see what CI says

https://github.com/ankitects/anki/releases/tag/24.10beta1


mikehardy:
per discussion on discord I will merge this and release it so it may be used as a regular dependency

*this is speculative*, it may be helpful, it may not.

It is likely not the only thing necessary to use anki 24.10 so I expect more commits to main will be necessary before release

If for some reason we need to use the old anki and do a release, the "fixup" for this untested "let's release a beta" process will be to go back to the commit before this one, create a git branch for stable releases, then release a version off that branch.

PR #417, opened 2024-10-04
Fix build

### LINKED ISSUE #416: Build is broken
Nope, repeat of same errors. Bizarre. This isn't pressing or anything, but might be some fun otherwise-useless CI+build fact we're all about to learn

_Originally posted by @mikehardy in https://github.com/ankidroid/Anki-Android-Backend/issues/415#issuecomment-2389423530_

dae:
- FSRS in Cargo.toml is ^1.2.2
- @l-m-sherlock made breaking changes without bumping major version number, and I didn't notice when reviewing the PR.
- AnkiDroid has its own lock file, so when the PR was made, a newer FSRS version got pulled in

Solved by making the requirement =1.2.2 and regenerating lockfile. We should probably use = in Cargo.toml moving forward - even minor fixes made to FSRS should be an explicit change in Anki, rather than being implicitly pulled in.

Closes #416

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "adler2"
+version = "2.0.0"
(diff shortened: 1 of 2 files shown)


mikehardy:
huzzah, thanks @dae

mikehardy (approved):
pending CI

PR #419, opened 2024-10-04
fix: adopt anki 24.10beta2

mikehardy:
bumped the local version since there were some changes here, as well as the anki part of version

we already published with 24.10beta1 so seems uncontroversial to publish with beta2 as well

it did cause a problem but was also one that seemed useful to see - a semver-minor breaking change, so this "is publishing a beta of anki useful?" experiment is still undecided but is not a fail yet and may be a positive as long it doesn't blow up in some non-useful way prior to stable release

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
 dependencies = [
  "gimli",
 ]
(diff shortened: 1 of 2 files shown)


mikehardy:
My "this doesn't seem controversial" stance implies that I'll just merge and publish as long as CI is good, can't see why issuing a fix release from upstream would be anything other than what we want

mikehardy:
Wut? This looks exactly like #416 meaning #417 didn't take for some reason? 🤔 main built after merging that though, and with a dependabot PR. I'm confused

dae:
#417 was a temporary fix. Short term fix for this PR: add = to 1.2.2 in cargo.toml and 'cargo check' to regenerate lockfile. Don't commit the toml change, as it will come in the next beta.

https://github.com/ankitects/anki/pull/3375/commits/f70e6e5695955545c135cbf12a51e06c2bc2bfc5

mikehardy:
My initial push following instructions obviously didn't work quite right and my immediate thought is "operator error" (that is, I just did it wrong). This is not a high priority and I value the education I'll get figuring it out so I'm still on this one and will get it working I think

mikehardy:
Decided on a different tactic: instead of hand-hacking local changes here, if it's fixed upstream already, just advance our commit hash here to the current one (which contains the fix)

So I bumped anki submodule to https://github.com/ankitects/anki/commit/407e2dc43b719f499d1d28f734e26620ce6a5fde

https://github.com/ankitects/anki/commits/main/

similar to "why not try releasing betas?" I don't see why - during a beta cycle upstream where we are *already* releasing betas - why we can't just advance slightly and go with it

Beats throwing away time on a local hand-hack dep pin IMHO

PR #420, opened 2024-10-10
build(ci): deflake steps with network fetches via retry

mikehardy:
Took a flake failure in CI last night and it so happens that the rust build also builds node things and that does network fetches

These are known flakes in CI, and handled well with a retry style, so wrapping the rust build in retries on all platforms in quick build now, and also in the build on release

Fail: 

```

rslib:i18n
failed: download:node
Error: error decoding response body

Caused by:
    0: error reading a body from connection
    1: end of file before message length reached
```

https://github.com/ankidroid/Anki-Android-Backend/actions/runs/11265493032/job/31327402861#step:13:459


mikehardy:
Note that the expansion in the release build from `build.sh` to `cargo run -p build_rust` is correct, based on casual inspection of contents of [build.sh](https://github.com/ankidroid/Anki-Android-Backend/blob/main/build.sh):

https://github.com/ankidroid/Anki-Android-Backend/blob/0b7fe85c13bb76a3c0ade1408ef5ef1b4eb9003b/build.sh#L1-L3

PR #422, opened 2024-10-10
fix(ci): required ubuntu 24 runner forward ports as `24` is now `latest`

mikehardy:
- with [much less default software](https://github.com/actions/runner-images/issues/10636) in the runner image, don't manually liberate space
- lzma no longer included, install ~~a package with stable name that will always depend on it ([xz-utils](https://packages.ubuntu.com/jammy/xz-utils))~~ the [liblzma-dev package](https://packages.ubuntu.com/noble/liblzma-dev) which should provide it

The need for this was noticed when a simple dependabot PR failed:

>   = note: /usr/bin/ld: cannot find -llzma: No such file or directory

https://github.com/ankidroid/Anki-Android-Backend/actions/runs/11278308705/job/31366466007?pr=421#step:13:40

...which led to a quick look at the runner image in use, noticed it was ubuntu-24 now and saw it just rolled out https://github.com/actions/runner-images/blob/ubuntu24/20241006.1/images/ubuntu/Ubuntu2404-Readme.md

Success criteria for merge is that the Ubuntu runner goes green in CI with runner image 24

PR #424, opened 2024-10-11
test(ci): pin ubuntu workflows to ubuntu-22.04 temporarily

mikehardy:
until we handle ubuntu-24 well, we need CI unblocked

If CI goes green, this is good to go

Related: #422 - most issues fixed there but still something wrong and I don't want it to be critical path when the pin workaround is easy


mikehardy:
This has already run successfully on the two affected workflows. Merging

PR #425, opened 2024-10-11
test(ci): use gradle directly vs deprecated gradle-build-action

mikehardy:
we have already converted to setup-gradle, which means that direct gradle usage should now be fully configured, and using gradle directly is the recommended style after using setup-gradle

success criteria is that the unit tests on all three build quick platforms run successfully, and that there are no longer deprecation warnings on the build-quick job summar


mikehardy:
success criteria met, just waiting for full green now just in case

PR #427, opened 2024-10-13
Add FSRS version to BuildConfig

voczi:
Excerpt from the Discord (slightly edited):
"i just used the FSRS crate version. actually pulling the version from the JS file (in the "normal" format, i.e. "FSRS 5.x.x") would require us to checkout the fsrs4anki repo with the branch==fsrs crate version, then pull the version from the js file with some regex.. very messy.
using the fsrs crate version would be that we can track any change to the FSRS crate, not just to the algorithm.
pulling the hash of the fsrs git commit can also be done, but is messy because anki is fed the fsrs dep through crates.io and not through github. so, best avoided. 
image"

Related issue: https://github.com/ankidroid/Anki-Android/issues/17236

### DIFF
--- a/build_rust/src/main.rs
+++ b/build_rust/src/main.rs
@@ -143,6 +143,7 @@ fn build_android_jni() -> Result<()> {
     let (is_release, _release_dir) = check_release(false);
 
     Command::run("cargo install cargo-ndk@3.5.4")?;
+    Command::run("cargo install jaq@1.6.0")?;
 
     let mut command = Command::new("cargo");
     command
--- a/rsdroid/build.gradle
+++ b/rsdroid/build.gradle
@@ -26,7 +26,7 @@ def getAnkiCommitHash = { ->
 def getFsrsVersion = { ->
     def pkgStdout = new ByteArrayOutputStream()
     exec {
-        commandLine "cargo", "metadata", "--format-version=1", "--manifest-path=" + new File("${project.rootDir}", "anki/Cargo.toml")
+        commandLine "cargo", "metadata", "--locked", "--format-version=1", "--manifest-path=" + new File("${project.rootDir}", "anki/Cargo.toml")
         standardOutput = pkgStdout
     }
 
--- a/rsdroid/build.gradle
+++ b/rsdroid/build.gradle
@@ -2,6 +2,7 @@ import com.android.build.gradle.tasks.BundleAar
 import com.vanniktech.maven.publish.SonatypeHost
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import java.util.zip.ZipFile
+import org.gradle.internal.os.OperatingSystem
 
 apply plugin: 'com.android.library' // required for aar generation to link to from AnkiDroid
 apply plugin: "kotlin-android"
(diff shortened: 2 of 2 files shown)


david-allison (dismissed):
Looks great, cheers!

voczi:
I'm just wrestling a bit with Gradle right now. Seems like JavaExec has different behavior for double quotes on different platforms..

voczi:
I'll see if I can avoid cargo downloading a bunch of crates to parse the version.. Otherwise, we will just use the "non-clean" route of grabbing it directly from Cargo.toml.

voczi:
Never mind. I don't think some extra crates being downloaded matters that much..
Maybe @mikehardy has an opinion here, but I think I prefer using the cargo command (+jaq to parse the JSON) over some hard-coded method.

Cargo.lock
```diff
@@ -1485,9 +1485,9 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "1.2.2"
+version = "1.3.1"
```
mikehardy, line 1488: why, on main, is Cargo.lock set with fsrs of 1.2.2 ? That pin should have moved with the 24.10beta2+ and beta3 submodule sha updates? 🤔 

Either way I need lock updates in separate commits because we have a release-2.19 branch here for me to release stable backend versions (that do *not* contain 24.10betas...) paired with the Anki-Android repo 2.19 release coming up

build_rust/src/main.rs
```diff
@@ -143,6 +143,7 @@ fn build_android_jni() -> Result<()> {
     let (is_release, _release_dir) = check_release(false);
 
     Command::run("cargo install cargo-ndk@3.5.4")?;
+    Command::run("cargo install jaq@1.6.0")?;
```
mikehardy, line 146: https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md#utilities

`jq` exists on the runner already why not use that? It works when I try it in buid.gradle

Oh, because then it's not cross-platform and it depends on local machine utility install status, whereas this just works. Nice job

Normally I'd just delete this comment but I thought you would find my thought process humorous, consider it a very tiny payment for the effort in making this PR

mikehardy (approved):
I like it, I'm going to dump the Cargo.lock change (consider in another PR that repairs whatever weirdness we have) though or at least separate it, then repush - so it's cherry-pickable

Cargo.lock
mikehardy: Okay, peeled this to #428 - it was my mistake originally, and you weren't wrong to have it in your local main but it was a whole process failure on my part. That PR should fix it, I'm purging it from this PR

mikehardy:
Pending CI, I expect to see an FSRS version 1.2.1 in there indicating this is running well
It worked great locally

mikehardy:
post-merge:
1- cherry-pick -x to release-2.19 branch
2- bump the gradle.properties local backend version number there
3- run release workflow with shipit input param *on the release-2.19 branch*
4- integrate this in to Anki-Android via a PR there

optional:
5- bump the gradle.properties local backend version *on main*
6- run release workflow with shipit input param *on main* for use in other PRs

...or could wait until next anki beta with FSRS v3/v4 cross-compat updates upstream

build_rust/src/main.rs
david-allison: The thought process would be great as a code comment

mikehardy: sure - good point - as long as I'm just rudely pushing to voczi/main from the PR (dude, I'm sorry - presumptuous I know 🫣 ) I went ahead and added that (then re-tested that I added a comment correctly, who knows) and re-pushed

rsdroid/build.gradle
```diff
@@ -22,6 +23,31 @@ def getAnkiCommitHash = { ->
...
+
+    def verArgs = OperatingSystem.current() == OperatingSystem.WINDOWS ?
+    ".packages[] | select(.name==\\\"fsrs\\\") | .version" :
+    ".packages[] | select(.name==\"fsrs\") | .version"
+    def verStdout = new ByteArrayOutputStream()
+    def verStdin = new ByteArrayInputStream(pkgStdout.toByteArray())
+    exec {
+        // use "jaq" cargo module installed during rust build: self-contained + cross-platform
+        // if we use `jq` we are dependent on local system utility installation status
+        commandLine "jaq", verArgs
```
mikehardy, line 41: comment felt more appropriate right here where it is used since that was when I went "wut?"

david-allison (approved):
LGTM, thanks so much!

@mikehardy leaving this one to you, as there are follow-up tasks

rsdroid/build.gradle
```diff
@@ -26,7 +26,7 @@ def getAnkiCommitHash = { ->
 def getFsrsVersion = { ->
     def pkgStdout = new ByteArrayOutputStream()
     exec {
-        commandLine "cargo", "metadata", "--format-version=1", "--manifest-path=" + new File("${project.rootDir}", "anki/Cargo.toml")
+        commandLine "cargo", "metadata", "--locked", "--format-version=1", "--manifest-path=" + new File("${project.rootDir}", "anki/Cargo.toml")
```
mikehardy, line 29: neat trick, wasn't aware of this one

PR #428, opened 2024-10-13
build(deps): result of `cargo check`, the FSRS update from anki 24.10beta3

mikehardy:
this should have been done and committed when updating the anki submodule sha

updated the documentation around bumping the anki sha as a note to my future self so that I will hopefully remember this step

Noticed this as I was reviewing #427 - because it had a lockfile FSRS bump in there when it really should not have, but the bump it had was important and should have already been done

Peeling it off here so that #427 is cherry-pickable to stable release-2.19 branch

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1485,9 +1485,9 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "1.2.2"
+version = "1.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


david-allison (approved):
Eventually we can verify this with CI

david-allison:
@mikehardy FYI: You're not proposing PRs from a fork of the repo

mikehardy:
I know that, I behave differently in thus repo since I'm only ever doing janitor work here - I just do it in repo for the most part. Can change if bothersome, certainly

david-allison:
No issue, as long as you're aware

PR #434, opened 2024-11-03
Include list of TS backend calls in artifacts directory

voczi:
Currently we have no way of knowing what backend functions we should support for the TS scripts in the Anki backend, beyond manually checking the code. This PR adds a list of function names that we should have bindings for in [PostRequestHandler.kt](https://github.com/ankidroid/Anki-Android/blob/main/AnkiDroid/src/main/java/com/ichi2/anki/pages/PostRequestHandler.kt), to the Anki artifacts directory in a file named "ts_funcs.txt". In the Anki-Android repo, we should add a test to verify that the keys of our mappings in PostRequestHandler correlate to the list in the artifact directory.
Also, we can use the crates from this PR (oxc_*) for other TS-related things. Considering the fact that the Anki ecosystem is moving to TS+Svelte, it could very well come in handy in the future.

### DIFF
--- a/build_rust/src/main.rs
+++ b/build_rust/src/main.rs
@@ -1,13 +1,16 @@
 use anki_io::{copy_file, create_dir_all};
+use anki_io::{create_file, read_file};
 use anki_process::CommandExt;
 use anyhow::Result;
 use camino::{Utf8Path, Utf8PathBuf};
 use std::env;
 use std::env::consts::OS;
 use std::fs;
-use std::io;
+use std::io::{self, Write};
 use std::path::Path;
 use std::process::Command;
+use streaming_iterator::StreamingIterator;
+use tree_sitter::{Parser, Query, QueryCursor};
 
 const ANDROID_OUT_DIR: &str = "rsdroid/build/generated/jniLibs";
 const ROBOLECTRIC_OUT_DIR: &str = "rsdroid-testing/build/generated/jniLibs";
@@ -86,6 +89,32 @@ fn build_web_artifacts() -> Result<()> {
 
     create_dir_all(artifacts_dir.join("js"))?;
     create_dir_all(artifacts_dir.join("css"))?;
+
+    let query = String::from_utf8(
+        read_file(Path::new("build_rust/tree_sitter_queries/ts_imports.scm")).unwrap(),
+    )
+    .unwrap();
+    let mut ts_funcs = vec![];
+    let mut ts_code = vec![];
+    ts_code.append(&mut ts_all_scripts(Path::new("anki/ts")).unwrap());
+    // Also include scripts inside of Svelte files.
+    ts_code.append(&mut svelte_all_scripts(Path::new("anki/ts/routes")).unwrap());
+    for script_code in ts_code {
+        // Get all of the imported backend funcs from the code.
+        ts_funcs.append(&mut ts_imported_funcs(script_code, &query).unwrap());
+    }
+
+    if let Ok(mut funcs_file) = create_file(artifacts_dir.join("ts_funcs.txt")) {
+        let mut content = ts_funcs
+            .iter()
+            .map(|name| name.to_string())
+            .collect::<Vec<String>>();
+        // Make sure the file ends with a newline.
+        content.push(String::new());
+        content.dedup();
+        funcs_file.write_all(content.join("\n").as_bytes()).unwrap();
+    }
+
     copy_file(
         "anki/out/ts/reviewer/reviewer_extras_bundle.js",
         artifacts_dir.join("js/reviewer_extras_bundle.js"),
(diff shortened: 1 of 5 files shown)


voczi:
Need to account for scripts inside Svelte pages, too?
https://github.com/ankitects/anki/blob/8b84583433d6400e64359dccfe1e2edde3df3e34/ts/routes/import-page/ImportPage.svelte#L13

mikehardy:
Yeah, `importDone` looks like a real symbol, perhaps ignore `pb` things in generated as that's protobuf? But include the other `@generated` items. 

I like where this is going though, it's similar to our Activity startup test. It was built to reproduce a real bug and has caught bugs before via this sort of "automated checking of universe of all possibilities" approach that scans for list of "should work" items then verifies they're handled...

voczi:
Considering another approach: https://tree-sitter.github.io/tree-sitter/
**Very** interesting project. Look at the list of parsers: https://github.com/tree-sitter/tree-sitter/wiki/List-of-parsers

mikehardy:
That looks pretty incredible yeah, drive-able in rust (apparently), can get AST of svelte or typescript? Appears to fit the bill
Between this and the PR where you took my grep and said "nope" then replaced it with proper XML parsing I'm starting to think you are a real programming language academic-type nerd! 👍 . Can't argue with it, this looks like the right sort of approach

voczi:
`ts_imported_funcs` is a mess. But.. at least it works? Or something? I don't know.
Here's how ts_funcs.txt looks like right now:
```
i18nResources
latestProgress
getSchedulingStatesWithContext
setSchedulingStates
cardStats
getChangeNotetypeInfo
getNotetypeNames
changeNotetype
getChangeNotetypeInfo
getNotetypeNames
congratsInfo
getDeckConfigsForUpdate
updateDeckConfigs
getDeckConfigsForUpdate
addImageOcclusionNote
updateImageOcclusionNote
getImageForOcclusion
getImageOcclusionNote
getImportAnkiPackagePresets
getCsvMetadata
getDeckNames
getNotetypeNames
getCsvMetadata
getFieldNames
importCsv
getCsvMetadata
getDeckNames
getNotetypeNames
importJsonFile
importJsonString
searchInBrowser
computeFsrsParams
computeOptimalRetention
simulateFsrsReview
evaluateParams
setWantsAbort
getGraphPreferences
graphs
setGraphPreferences
importAnkiPackage
importDone
```

voczi:
Useful for future work: https://tree-sitter.github.io/tree-sitter/playground

build_rust/src/main.rs
```diff
@@ -317,3 +345,93 @@ fn copy_dir_all(src: impl AsRef<Path>, dst: impl AsRef<Path>) -> io::Result<()>
...
+    let query = Query::new(&tree_sitter_typescript::LANGUAGE_TYPESCRIPT.into(), query).unwrap();
+
+    let mut imports = vec![];
+    let mut matches = query_cursor.matches(&query, tree.root_node(), code_bytes);
+    while let Some(item) = matches.next() {
+        let captures = item.captures;
+        // The name of an import. E.g. "importDone" if the import line is:
+        // "import { importDone } from "@generated/backend";
+        // Should always be the first captured field.
+        let name = captures.get(0).unwrap().node.utf8_text(code_bytes).unwrap();
```
mikehardy, line 425: Hey @voczi - I'm unsure exactly how this works in the anki-upstream / Anki-Android-Backend context, but I know in typescript it is possible to have multiple comma-separated named imports, but if I am reading this correctly only the first named import is seen here?

build_rust/src/main.rs
```diff
@@ -86,6 +89,31 @@ fn build_web_artifacts() -> Result<()> {
 
     create_dir_all(artifacts_dir.join("js"))?;
     create_dir_all(artifacts_dir.join("css"))?;
+
+    let query =
+        String::from_utf8(read_file(Path::new("build_rust/queries/ts_imports.scm")).unwrap())
```
mikehardy, line 94: I do a lot of directory tree browsing and if I see just "queries" I'll think it's generic but we're just storing tree sitter queries, propose directory rename queries -> tree_sitter_queries

Also, all bike sheds should have murals painted on them by local graffiti artists, funded by city grants, but only by those artists that promise not to tag historical buildings.

mikehardy (changes requested):
looks pretty good, couple comments

needs a rebase to main after anki24.10rc2 merges, and should contain a gradle.properties bump to the local part of the version number, as you're adding a cool feature here so it should be a distinct version

build_rust/src/main.rs
voczi: I assume that every unique pair of captures is iterated here. So if you have something like
`import { importDone, doXYZ } from "@generated/backend";`
Then you get a match where name="importDone"&src="@generated/backend" and another where name="doXYZ"&src="@generated/backend".
I should double-check this.

build_rust/src/main.rs
voczi: > I do a lot of directory tree browsing and if I see just "queries" I'll think it's generic but we're just storing tree sitter queries, propose directory rename queries -> tree_sitter_queries

Sure!

> Also, all bike sheds should have murals painted on them by local graffiti artists, funded by city grants, but only by those artists that promise not to tag historical buildings.

WHAT

mikehardy: I am the GOAT of bike shed color choosers and just wanted to demonstrate my unmatchable power

Managed to delegate color choice to something that can't be judged (it's art!) with decision process speed determined by a government committee. Beat that.

voczi: LGTM 🚀

build_rust/src/main.rs
mikehardy: if you double-checked and it's good I'm +1 for ship

PR #437, opened 2024-11-06
Increase scope of TS code scraping

voczi:


mikehardy:
The build/release here is pretty resource intensive - not just in CI but the artifacts are pretty big on they're on maven forever
Have you tested this locally by pointing Anki-Android repo to it using gradle.properties to use local backend, and it's definitely working ?

Just want to make sure we're not iterating here as if the process was lightweight because it is definitely not...

voczi:
> Have you tested this locally by pointing Anki-Android repo to it using gradle.properties to use local backend

Doing it right now actually. See no reason for it not to work, but I'll just report back whenever it's built.

mikehardy (approved):
LGTM

PR #448, opened 2024-11-26
fix: adopt anki 24.11rc2

mikehardy:
built cleanly but only I after I did this locally:

> cd anki && git clean -d -x -f

Before that there was some issue with (apparently cached?) `portal.ts` in svelte area, which has been moved to a component

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -19,7 +19,7 @@ android.useAndroidX=true
 android.enableJetifier=false
 
 GROUP=io.github.david-allison
-VERSION_NAME=0.1.47-anki24.11rc1
+VERSION_NAME=0.1.47-anki24.11rc2
 
 POM_INCEPTION_YEAR=2020
 


gradle.properties
```diff
@@ -19,7 +19,7 @@ android.useAndroidX=true
 android.enableJetifier=false
 
 GROUP=io.github.david-allison
-VERSION_NAME=0.1.47-anki24.11rc1
```
david-allison, line 22: We've done an Anki bump in the past without a version number, but does this warrant `0.1.48`?

Regardless of the outcome. Approved

david-allison (approved):

gradle.properties
mikehardy: No it doesn't - we've already done a local version bump for the changes required here so to me this is anki upstream only, there were no changes here
Not that it's truly important, just that's how I reason about it

david-allison:
Released: https://github.com/ankidroid/Anki-Android-Backend/actions/runs/12050946050

PR #457, opened 2025-01-20
feat: adopt anki25.01beta1

mikehardy:
This had local build failures for me on macOS due to a missing `cairo/cairo.h` header file included in the node dependency `canvas`.

I actually have cairo installed on my mac and I made no good headway within my timebox to figure out why it wasn't being seen. I do have a fairly custom node setup locally though, and that *may* be the culprit

So, I'm tossing it into CI to see what happens there in stock setups of everything

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1489,9 +1489,9 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "1.4.3"
+version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "de570dbbc7b95c5cf5f77c905b3b54b16352c1b48694e6761d716fb92b8db67c"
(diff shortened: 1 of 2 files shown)


iamllama:
Related to https://github.com/ankitects/anki/issues/3635?

mikehardy:
@iamllama well-spotted, indeed I have been slowly migrating my personal projects to yarn berry - and I suspected this was the cause of my issues which is why I posted the PR to see how a yarn classic environment handled it

Built fine in CI, so I think yes - I was bitten by that issue. Thanks for pointing it out!

Would be better if it worked with modern yarn though, IMHO :-), I like to fix things forward if possible. But also I pin things to older versions temporarily to get things moving... so the suggested pin of yarn to yarn classic seems practical

We're unblocked here though, I can merge this since our CI build uses yarn classic

PR #462, opened 2025-01-27
build: add diagnostics for windows jaq install hang

mikehardy:
Something is causing windows build to hang on the jaq crate install, this should help?

There is some chatter about sub-processes hanging just by not having their stdout/stderr consumed in windows, so I may affect the hang (and effectively solve it) simply now consuming the output. But if not then we'll have the output to examine anyway

Also updates jaq to 2.1.0, which may also inadvertently solve it, because perhaps jaq was already installed but an older version before - so install never really happened so there was no problem with the install - but now that 2.1.0 is out an 1.6.0 is maybe not installed the nstall was happening and causing the problem

All  know is I want the windows build in CI to work again. It was working for me in windows when I tried to reproduce, so I'm not really sure what's going on

### DIFF
--- a/.github/workflows/build-quick.yml
+++ b/.github/workflows/build-quick.yml
@@ -9,13 +9,16 @@ on:
   push:
     branches:
       - main
+  merge_group:
 
 env:
   DEBUG: 1
   ANDROID_NDK_VERSION: "27.0.12077973"
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  # if a workflow is run against the same ref, only one at a time...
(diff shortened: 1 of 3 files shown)


mikehardy:
This appears to fix the windows build hang. Not sure which part of it was "the" fix, but it's working now where it was consistently failing before. And at minimum it is an improvement in that we have output now and we have updated jaq now

david-allison (approved):

mikehardy:
this includes a change which should be last part to getting merge queue working, but those are workflow changes which won't be on main

So I'm going to slam this one in, then rebase the kotlin bump to it and see if that sails through with queue working fine

PR #463, opened 2025-02-11
Anki25.02

mikehardy:
- update local java dependencies (need to examine why dependabot didn't get them)
- fix new gradle "space assignment is deprecated" warnings in local gradle build files
- adopted anki 25.02
  - updated local rust-toolchain.toml and added that step to docs

Let's go


mikehardy:
> Commit 1 may also warrant automation investigation

I investigated our dependabot.yml and the settings and...I dunno I don't see anything amiss. And it did PR the kotlin dep recently ? I just don't know. Maybe they were ignored somehow and that stuck in dependabot status somewhere. Will look one more time 🤷 . IDE flags them at least

PR #464, opened 2025-02-11
build: convert settings.gradle to Kotlin DSL

david-allison:
We were having problems with dependabot not detecting dependencies in:

* #463

This only affected dependencies in subprojects

This is an attempt to fix it


mikehardy (approved):
🤔  can't hurt, .kts is the future anyway, but I'll be surprised if this was it!

PR #471, opened 2025-02-15
fix(deps): define NDK version in a single spot, use dynamically everywhere

### LINKED ISSUE #467: Put NDK version in libs.versions.toml and reference it dynamically
Right now the NDK version is scattered everywhere, it should be in one spot

There was a previous issue #405 but I closed that in order to honor wishes of a previous contributor

Goal here is to have all versions in one file

mikehardy:
Previously the NDK version we wanted to use was scattered in 14 spots.

One of the more thorough violations of the principal of "Don't Repeat Yourself" I've ever seen

- Fixes #467 

Once I get this through I'll do a follow-on where I try to adopt the newer v28 release of the ndk

### DIFF
--- a/.github/workflows/build-quick.yml
+++ b/.github/workflows/build-quick.yml
@@ -13,7 +13,6 @@ on:
 
 env:
   DEBUG: 1
-  ANDROID_NDK_VERSION: "27.0.12077973"
 
 concurrency:
   # if a workflow is run against the same ref, only one at a time...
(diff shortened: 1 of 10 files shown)


mikehardy:
Sweet, worked perfectly first time (lying: took a while locally, with many failures, but working on first push to github does feel nice)

david-allison (approved):
Much better than my initial attempt, cheers!

PR #472, opened 2025-02-15
test(ci): use !cancelled instead of always, avoid missing emu artifacts

mikehardy:
- always() status check also runs when runs are cancelled, but you don't want that, so it is recommended to use '!cancelled()' as an "always, unless cancelled"
- we were trying to upload emulater log / video on all platforms but it only runs on ubuntu

Testing:
If this is correct, we should see 
- 2 rsdroid artifacts for each of the three platforms, and 
- 2 emulator artifacts for ubuntu, and no artifact upload warnings


mikehardy:
looks good!

PR #473, opened 2025-02-15
build(deps): adopt ndk v28 / add doc on CLI install of same

mikehardy:
does what the title says - ndk28 + a quick doc for CLI folks to install it quickly

works locally with Anki-Android via `local_backend=true`, everything *seems* fine

super easy now that #471 is in

### DIFF
--- a/rslib-bridge/build.rs
+++ b/rslib-bridge/build.rs
@@ -7,6 +7,7 @@ pub mod proto;
 use std::{
     env::{self, consts::OS},
     path::PathBuf,
+    process::Command,
 };
 
 use anyhow::Result;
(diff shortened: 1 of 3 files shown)


mikehardy:
Ha, of course it worked in CI on macOS when compiling for arm because that's what I tested locally and it "Just Worked(tm)"...but then fails on all other platform combinations.

Okay, more testing to do here clearly.

mikehardy:
Apparently there is a very very very subtle rust upstream issue that we have a linker workaround for, but our workaround was hard-coding the `clang` version it expected to find when generating linker workaround paths.

That is fragile vs the NDK as the clang version changes over time.

I found an updated version of the workaround that dynamically fetches the clang version for use in the linker workaround path generation, and I've incorporated it here

rslib-bridge/build.rs
```diff
@@ -33,10 +34,41 @@ fn main() -> Result<()> {
         } else {
             "windows-x86_64"
         };
-        let lib_dir = format!("/toolchains/llvm/prebuilt/{platform}/lib/clang/18/lib/linux/");
```
mikehardy, line 36: The '18' in that path was the problem.

All the `+` lines here are just to dynamically find clang, run it to get the current version (maybe not 18 anymore...) and use that instead of a hard-coded 18.

All makes sense when you look at it

david-allison (approved):
LGTM

PR #495, opened 2025-03-25
build(deps): cargo update / cargo upgrade / recreate Cargo.lock

mikehardy:
Still fiddling around with getting dependencies truly up to date

Not the highest value work but I'm learning a lot about the Rust toolchain in the process, which is why I'm doing it
Strangely, I always start a language by really trying to understand the tools and automate basic workflows

- `cargo upgrade` comes from the cargo-edit package
- had to force one dep down that wasn't compatible with 1.84 rust

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -26,7 +26,7 @@ dependencies = [
  "cfg-if",
  "once_cell",
  "version_check",
- "zerocopy",
+ "zerocopy 0.7.35",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


david-allison (approved):

mikehardy:
That seemed to get us to a usable baseline. Green dependabot for cargo now https://github.com/ankidroid/Anki-Android-Backend/actions/runs/14070097108/job/39402063934

PR #496, opened 2025-03-25
build(deps): adopt ndk29

mikehardy:
note, this is not ready yet, but a final commit when it is released as stable should fix it up, and it already appears to work fine

Just posting it because I was in there and gave it a shot and it worked, posting it up to show this is all NDK bump takes now

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -4,7 +4,7 @@ targetSdk = "34"
 minSdk = "21"
 
 # https://developer.android.com/ndk/downloads
-ndk = "28.0.13004108"
+ndk = "29.0.13113456"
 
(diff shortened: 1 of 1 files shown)


mikehardy:
Oh that's interesting:

> Accept? (y/N): Skipping following packages as the license is not accepted:

So if it is not already installed, our current installation process / old-version cleaner for NDK does not work since it does not accept the license.

Noted. Not a priority to fix but I'll probably swing by and get it at some point

mikehardy:
Okay, this is now passing CI and just waiting for upstream to release as stable.
Not important, but pre-qualifying dependencies is interesting when I have time

david-allison (approved):

PR #499, opened 2025-04-02
build: pre-test on windows-2025

mikehardy:
This is not an important PR, was just clearing notification queue and saw that windows-2025 was available as a runner

This PR is nothing more than a compatibility check. If it runs fine I will close it rather than put the change in as I prefer `-latest` tags, but if it fails to run for some reason then it bears investigation so we're ready whenever `windows-latest` moves to the windows-2025 runner


mikehardy:
Excellent, no issues. I'll close this - and rest easy that when windows-latest tag slides to windows-2025 we'll be fine.

PR #505, opened 2025-04-27
Properly format generated kotlin code from proto

Arthur-Milchior:
Tested by looking at GeneratedBackend.kt before and after the change, and seeing something that now looks like what kotlin is supposed to look like.


mikehardy (approved):
:-) works for me - I agree with you, I think even generated code should look correct, it's automated after all. For that reason especially, there's no reason it should look bad

Arthur-Milchior:
Thank you for merging it @mikehardy. It'll reduce frustration so much when I look at the generated code!

PR #506, opened 2025-04-27
Add pre-commit format-rust

Arthur-Milchior:
I've not found an already existing script to format the kotlin part. So I've only added the rust check right now. That would have avoided the delay in #505


mikehardy (approved):
LGTM - thanks

PR #514, opened 2025-05-15
fix(deps): adopt upstream anki 25.02.5

mikehardy:
https://github.com/ankitects/anki/releases/tag/25.02.5

I don't think any of this affects us, but best to be current in case next one does


mikehardy:
Just kicked off build release workflow in publish mode for this one, ETA hour or two before it is up on maven central

PR #525, opened 2025-06-10
build(deps): adopt anki 25.06b1

mikehardy:
Initial attempt to build out anki 25.06 - working okay?

# issue

some build artifact that requires `protoc` is now required by `cargo check` ?

You used to be able to run `cargo check` before running `cargo run -p build_rust` (which has a build.sh or build.bat shortcut for running same)

It will error now requiring protoc (which is downloaded internally by some build step) and a descriptors.bin file that isn't generated but needed by some other build step

✅  if you just do the build before doing the cargo check, it works fine though

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -112,7 +112,7 @@ dependencies = [
  "csv",
  "data-encoding",
  "difflib",
- "dirs",
+ "dirs 5.0.1",
  "envy",
  "flate2",
  "fluent",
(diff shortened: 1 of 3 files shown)


david-allison:
@mikehardy This will block 2.21 on a stable release of 2025.6

Right decision IMO, but to be sure you're aware

mikehardy:
Going to publish this now - note for the avoidance of doubt that the intention is for easy integration feasibility testing on AnkiDroid, not to actually merge it over there (yet)

Should we need to merge any changes here on Backend for use over there before 25.06 is released stable and integrated, the idea is to retro-branch here in Backend just prior to this merge, and implement + publish changes as needed.

I doubt that will be necessary but it's the plan if needed.

PR #528, opened 2025-06-14
feat: log panics as 'Error' in logcat

david-allison:
Previously these logs were 'debug', and therefore not in ACRA

* For https://github.com/ankidroid/Anki-Android/issues/18512

----

@dae requesting review 

* Could you confirm we're not going to receive email addresses in the panics 
  * if so, is there a good method of avoiding this. I believe we already filter server-side, but ideally don't want to transmit them at all
* (optional) A quick check of the Rust if you have time

----

* @mikehardy could I confirm we'd handle PII correctly if it was passed in via this PR
* https://github.com/ankidroid/Anki-Android/issues/17392

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -104,7 +104,7 @@ jobs:
           max_attempts: 3
           retry_wait_seconds: 0
           retry_on: error
-          command: cargo run -p build_rust
+          command: ./build.sh
 
       - name: Upload rsdroid AAR as artifact
         uses: actions/upload-artifact@v4
--- a/README.md
+++ b/README.md
@@ -29,6 +29,16 @@ Install Xcode/Visual Studio if on macOS/Windows.
 
 Install rustup from <https://rustup.rs/>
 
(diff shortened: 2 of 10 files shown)


david-allison:
> Assuming the Rust code works, I can't see any obvious issues with it. That said, there's some panic handling in panic_to_backend_error that I assume is supposed to be already handling this, so I'm mildly curious about the root cause of these panics. Maybe they're happening in a thread we spawn for media syncing or similar, so they're not caught by a direct invocation.

Thanks! I also suspect it's background media syncing related. We'll know shortly.

david-allison:
force pushed:

* fixed clippy issue
* proposed pre-commit hook: #529
* bumped version

```diff
- 0.1.52-anki25.06b1
+ 0.1.53-anki25.06b1
```

mikehardy:
fantastic - any related PR on the Anki-Android side should just be forced through IMHO, if I don't see it. No reason to wait IMHO, faster we can get a build out that exposes the underlying backtrace that's causing the poison error, the better

PR #534, opened 2025-06-20
build(deps): adopt anki 25.06b5

david-allison:
⚠️: I was warned we're in a detached HEAD state with Anki, @mikehardy this isn't documented.

I'm using the correct commit for https://github.com/ankitects/anki/releases/tag/25.06b4

----

Had the following temporary error when running `./build.sh` Worked again after rebuild

```
     Running `out/rust/debug/configure`
n2: error: /Users/davidallison/StudioProjects/Anki-Android-Backend/anki/out/build.ninja:135: input ftl/core-repo/core/pt-PT/network.ftl missing
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -46,13 +46,13 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "ammonia"
-version = "4.0.0"
+version = "4.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1ab99eae5ee58501ab236beb6f20f6ca39be615267b014899c89b2f0bc18a459"
+checksum = "3ada2ee439075a3e70b6992fce18ac4e407cd05aea9ca3f75d2c0b0c20bbb364"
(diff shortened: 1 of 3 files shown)


david-allison:
👀

```
net.ankiweb.rsdroid.database.CorruptDiskFullDatabaseTest > testCorruption[RUST][emulator-5554 - 6.0] FAILED 
	java.lang.AssertionError:
	Expected: type < android.database.sqlite.SQLiteFullException
```

mikehardy:
running `check-rush.sh` results in an updated Cargo.lock which makes sense as `nom` was updated in anki upstream - we should have that locked here. Adding it to the commit...

david-allison:
I'm happy to ignore the test. It was designed to be sure that the Rust Backend threw the same exceptions as the Android Framework `SupportSQLiteDatabase`. 

Since we're effectively fully migrated to the new backend now, maintaining an exact 1:1 mapping with Android isn't necessary, especially as this error likely occurs when loading the collection

david-allison:
https://github.com/ankidroid/Anki-Android-Backend/actions/runs/15809266570

Issue #70, opened 2021-04-04
Add auto-formatting to build pipeline


david-allison-1:
- [ ] rustfmt
- [ ] Some Java formatting (Google?)

mikehardy:
This is interesting - some other repos I work on have cast about for something for java

https://github.com/google/google-java-format

And their release notes may incidentally fix the AnkiDroid JDK16 build issue https://github.com/ankidroid/Anki-Android/issues/8434 !

https://github.com/google/google-java-format#jdk-16

mikehardy:
this is in actual use in a repo I maintain with others, and thus maybe useful as a recipe

>    "format:android": "google-java-format --replace -i $(find . -type f -name \"*.java\")",

dae:
Rust formatting is already covered by CI / check-rust.sh. Kotlin formatting could presumably be handled in a similar way Anki-Android.

PR #543, opened 2025-06-25
Let's get ktlint going in this repo

mikehardy:
- port the main build file to kotlin so we can share tricks from Anki-Android repo
- allow jdk24 because I'm qualifying it locally, using the tricks from Anki-Android
- put some exclusions useful for this repo into .editorconfig that make lint more palatable (that is: everything autocorrected, no manual intervention needed)
- run ktlint
- also add in gradle plugin while I'm here

Related, possible now, but not fixed quite yet:

- #70


mikehardy:
> Great catch, cheers!

Yeah not super high value but prompted me to start kotlin build and lint, so, some background progress at least...

PR #544, opened 2025-06-28
build(deps): adopt upstream anki 25.06b6

mikehardy:
note, rust-toolchain bumped upstream, I simply mirror that here

ran ./check-rust.sh which sometimes bumps things in our Cargo.lock but it did no such thing, so this should be pretty simple and good to go once CI approves


david-allison:
commit hash is correct: https://github.com/ankitects/anki/releases/tag/25.06b6

LGTM if CI is happy

PR #550, opened 2025-07-12
build(deps): adopt anki upstream 25.07.2

mikehardy:
Not sure whether we want 25.08b1 or 25.07.2 so posting them both


mikehardy:
cross-link with related optional bump:

- #551

mikehardy:
re-pushed with result of NDK merge:

- #509 

Will set this to merge and publish it, then integrate that in main on Anki-Android per discussion

PR #576, opened 2025-09-12
build(deps): update to minSdk 23

david-allison:
required by androidx.sqlite 2.6.0

> Updated the library’s Android minSDK from API 21 to API 23.

https://developer.android.com/jetpack/androidx/releases/sqlite#2.6.0

* Unblocks https://github.com/ankidroid/Anki-Android-Backend/pull/575

----

`useLegacyPackaging` was required, otherwise android tests failed with

`dlopen failed: library "librsdroid.so" not found`

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 compileSdk = "34"
 targetSdk = "34"
-minSdk = "21"
+minSdk = "23"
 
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
seems fine - and the legacy packaging thing also seems fine until they drop support for it, unless someone wants to take it up I think it only affects performance for androidTest uses, which is dev-only in Anki-Android repo

PR #586, opened 2025-10-16
fix(deps): adopt anki upstream v25.09.2

mikehardy:
Note: used internal version 0.1.62 vs 0.1.61 because there are a number of other dependency bumps landing here at the same time, enough so it seemed a good idea to bump internal (non-anki) portion of the version segment in prep for a publish with all changes at once...

----

now `0.1.62-anki25.09.2`

https://github.com/ankitects/anki/releases/25.09.2/

hash: 3890e12c9e48c028c3f12aa58cb64bd9f8895e30

```
git fetch --all
cd anki
git fetch --all
git checkout 25.09.2--recurse-submodules
cd ../
./build.sh
./check-rust.sh
```

and update gradle.properties

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -19,7 +19,7 @@ android.useAndroidX=true
 android.enableJetifier=false
 
 GROUP=io.github.david-allison
-VERSION_NAME=0.1.61-anki25.09
+VERSION_NAME=0.1.62-anki25.09.2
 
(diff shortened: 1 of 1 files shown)


david-allison (approved):
Hashes match, definitely should be a version bump due to the dependency updates. Thanks!

PR #587, opened 2025-10-17
perf(ci): use self-hosted macos runner if available, fallback to public runners if not

mikehardy:
GitHub macos runners are performing incredibly badly at the moment

This is a change that lets us use Tartelet / Tart to do a self-hosted runner, with fallback if they're offline

https://josephduffy.co.uk/posts/self-hosting-macos-github-runners

- local app that does the VM clone/start/GH-register/destroy - https://github.com/shapehq/tartelet
- local VM system built on macOS virtualization - https://github.com/cirruslabs/tart

If this should be unwound:
- remove the github app for it in ankidroid org app settings
- remove the org level secret with token used to access the github API to query for runner status

### DIFF
--- a/.github/workflows/build-quick.yml
+++ b/.github/workflows/build-quick.yml
@@ -140,37 +140,6 @@ jobs:
       - name: Fetch submodules
         run: git submodule update --init
 
-      - name: Read configured NDK version
-        run: |
-          cargo install toml-cli
-          ANDROID_NDK_VERSION=$(toml get gradle/libs.versions.toml versions.ndk --raw)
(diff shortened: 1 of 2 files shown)


mikehardy:
Success on the first run, how about that.

Took 22mins to run the macos tests on build-quick workflow, of which 3'23" was restoring rust cache, which was a 2GB network transfer (the largest transfer in the entire workflow)

Going to eliminate the rust cache for self-hosted builds, and also alter the release workflow to use the macos-tartelet runner, and see how it goes

PR #595, opened 2025-11-04
refactor(deps): unify junit ref name

david-allison:
As-of JUnit 6, all packages share the same version scheme our ref name implies we didn't update

This wasn't a problem, but I had to check in a code review, and will have to check for future reviews

> Single version number for Platform, Jupiter, and Vintage

https://docs.junit.org/6.0.0/release-notes/#release-notes-6.0.0

### DIFF
--- a/.github/workflows/build-quick.yml
+++ b/.github/workflows/build-quick.yml
@@ -29,7 +29,7 @@ jobs:
     steps:
       - name: Use self-hosted runner if online and not busy, otherwise public runner
         id: determine-macos-runner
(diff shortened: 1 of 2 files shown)


mikehardy:
```

SyntaxError: Unexpected token 'm', "macos-26" is not valid JSON
Error: Unhandled error: SyntaxError: Unexpected token 'm', "macos-26" is not valid JSON
    at JSON.parse (<anonymous>)

```

That looks like it's "on me" as it were, I must have done something wrong in my quixotic self-hosted-runner adventure

I'll own this and get that fixed and the PR through

PR #601, opened 2025-11-14
fix(ci): runner determination needs to run serially

### LINKED ISSUE #600: [CI] The self-hosted runner lost communication with the server.
> The self-hosted runner lost communication with the server. Verify the machine is running and has a healthy network connection. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

https://github.com/ankidroid/Anki-Android-Backend/actions/runs/19377284899

mikehardy:
otherwise two workflows triggered in parallel from the same event could both determine that self-hosted is "available" and ther are enough instances available, even though only one runner is available, because they both run the determination before the runner is occupied by the later jobs in the same workflow

- Fixes #600 

(...which only happened because I was manually / contemporaneously watching the underlying issue this solves cause unexpected behavior on the computer next to me and in workflow sequencing, and I killed the runner while it was in process...)

### DIFF
--- a/.github/workflows/build-quick.yml
+++ b/.github/workflows/build-quick.yml
@@ -24,6 +24,9 @@ jobs:
   # We may have a self-hosted runner available for macOS. Use it if so.
   determine-macos-runner:
     runs-on: ubuntu-latest
+    concurrency:
+      group: runner-determination
+      cancel-in-progress: false
     outputs:
       runner: ${{ steps.determine-macos-runner.outputs.use-runner }}
     steps:
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -34,6 +34,9 @@ jobs:
   # We may have a self-hosted runner available for macOS. Use it if so.
   determine-macos-runner:
     runs-on: ubuntu-latest
+    concurrency:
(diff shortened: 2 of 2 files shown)


mikehardy:
My initial test here wasn't great - I'll appreciate the review but please don't merge yet, I have 2 runners up so they both were selected as there was zero contention. But I'm not sure if the concurrency would have prevented queueing if there was only one runner up. I'll need to kill jobs and change things locally and retry etc.

david-allison (approved):

mikehardy:
Okay, near as I can tell the runner *determination* step is running serially now - they each had timestamps in non-overlapping sequential blocks of time. So that's only one data point but it appears good and gives me a chance to manage the thundering herd issue.

However, an analysis of the merge (push to main) event's workflows indicates that the actual build job that occupies the runner took just long enough to kick off that the second runner determination still ran before the runner was occupied, leading to a job queueing on the self-hosted runner instead of falling back.

It was only a 2 second miss, timing wise. I'm going to make one more tweak - I'm going to add a 15sec sleep before the runner determination to give a chance for jobs initiated by the previous determination to impact system state.

PR #602, opened 2025-11-15
test(ci): wait for parallel runs to occupy runners before choosing

mikehardy:
Hopefully the last tweak here to have a system that meets the stated goal:

workflows here should never error or queue waiting for a self-hosted runner, regardless of self-hosted runner state

If it's totally down, github hosted runners should be used
if it is up but occupied, no jobs should queue, github hosted runners should be used

Sounds simple but has some subtlety about it apparently

### DIFF
--- a/.github/workflows/build-quick.yml
+++ b/.github/workflows/build-quick.yml
@@ -30,6 +30,10 @@ jobs:
     outputs:
       runner: ${{ steps.determine-macos-runner.outputs.use-runner }}
     steps:
+      - name: Wait for possible parallel workflow run job startup lag
+        # After runner choice, the job that will use it has unavoidable job startup lag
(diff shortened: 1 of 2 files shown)


mikehardy:
Perfection! I think.

- I can see that the concurrency control worked, and the web UI even helpfully told me why the second runner choice job was waiting
- I can see that with the added sleep, there was enough time for the previous choice to have the inevitable impact on system state of occupying the runner and the parallel workflow chose correctly

image

mikehardy:
Ah, just a bad coincidence. The laptop went to sleep while it was in the run. I don't think that's a solve-able problem except now I've set it to not sleep but turned the screen off. I'm okay with that - the runs are easily restartable anyway and that's just bad luck I think

PR #673, opened 2026-03-29
use local maven repo since linking to the .aar directly makes things not work

### LINKED ISSUE #671: can't compile ankidroid using local_backend
trying to compile ankidroid with `local_backend=true`:
```
1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':libanki:bundleDebugAar'.
> Error while evaluating property 'hasLocalAarDeps' of task ':libanki:bundleDebugAar'.
   > Direct local .aar file dependencies are not supported when building an AAR. The resulting AAR would be broken because the classes and Android resources from any local .aar file dependencies would not be packaged in the resulting AAR. Previous versions of the Android Gradle Plugin produce broken AARs in this case too (despite not throwing this error). The following direct local .aar file dependencies of the :libanki project caused this error: /home/jake/code/Anki-Android-Backend/rsdroid/build/outputs/aar/rsdroid-release.aar
```

using a local maven repo solves this

jakeprobst:
closes #671

disclaimer: I have no idea what I am doing. I have managed to get this all working but I am unsure if this is the _correct_ way to do things.

I don't know how things are published normally, but it is configured such that `publishReleasePublicationToMavenRepository` will work fine where `publishReleasePublicationToMavenCentral` requires some sort of credentials I do not have. so I am assuming is is correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,8 @@ var RELEASE=1 defined, to build a faster version.
 
 ## Modify AnkiDroid to use built library
 
+Install the .aar to a local maven repo with `./gradlew :rsdroid:publishReleasePublicationToMavenRepository`
+
 Now open the AnkiDroid project in AndroidStudio. To tell gradle to load the
 compiled .aar and .jar files from disk, edit `local.properties`
 in the AnkiDroid repo, and add the following line:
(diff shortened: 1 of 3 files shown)


jakeprobst:
it seems the preferred solution is to make it so `rsdroid` doesn't have any android deps and so is a normal package and not an android package

PR #713, opened 2026-08-09
feat(class-loader): handle backend requests from multiple classloaders

david-allison:
> [!NOTE] 
> Assisted-by: Claude Opus 5 - implementation

> [!NOTE]
> (cherry picked from commit cde8375625f0dc9e5e287aeedac992ff8c3736de) - the change should not have gone to `main` without review.

Each Classloader needs a separate on-disk file or a `UnsatisfiedLinkError` is thrown.

To fix this:

* Persist a collection of 10 files on-disk
  * We do this as we cannot guarantee file deletion on Windows

```
librsdroid-<sha1>.dylib      the extracted library
librsdroid-<sha1>-1.dylib    a copy of the library, created if a second classloader requires it
```

* Walk through the libraries using a cursor and load an unused one
* Devise a script for Anki-Android to warn developers if there are too many stale copies on disk (each is 60MB)

----

* Fixes most of the following, a merge of an updated backend library will resolve it ankidroid/Anki-Android#14796

### DIFF
--- a/rsdroid-testing/src/main/java/net/ankiweb/rsdroid/testing/FileUtils.kt
+++ b/rsdroid-testing/src/main/java/net/ankiweb/rsdroid/testing/FileUtils.kt
@@ -1,0 +1,35 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+package net.ankiweb.rsdroid.testing
+
+import java.io.File
+import java.io.IOException
+
+/**
+ * Copies this file to [destination] [atomically][createAtomically].
+ *
+ * @throws IOException [destination] could neither be created nor found
+ */
+internal fun File.copyAtomicallyTo(destination: File) {
+    destination.createAtomically { temporary -> copyTo(temporary, overwrite = true) }
+}
+
+/**
+ * Creates this file in an atomic manner.
+ *
+ * If another process creates this file first, their file is left in place.
+ *
+ * @param write populates the temporary file which becomes this file
+ * @throws IOException this file could neither be created nor found
+ */
+internal fun File.createAtomically(write: (File) -> Unit) {
+    val temporary = File.createTempFile("$nameWithoutExtension-", ".tmp", parentFile)
+    try {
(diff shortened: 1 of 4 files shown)


BrayanDSO:
I'm okay with the compromise.

Is the ideal number of copies 10? Could it be less or is it the best alternative?

david-allison:
10 exists to allow for some growth room. I'd be happy to halve it to 5.

I suspect in reality we'll be fine with 3 for a while, but this will need to be merged and implemented to know for certain

david-allison:
Halved to 5. Will squash when I'm back at a PC

rsdroid-testing/src/main/java/net/ankiweb/rsdroid/testing/FileUtils.kt
```diff
@@ -0,0 +1,21 @@
...
+ * so [destination] is never seen partially written by a process which opens it as it appears.
+ *
+ * If another process creates [destination] first, their file is left in place.
+ */
+internal fun File.copyAtomicallyTo(destination: File) {
+    val temporary = File.createTempFile("${destination.nameWithoutExtension}-", ".tmp", destination.parentFile)
+    try {
+        copyTo(temporary, overwrite = true)
+        // this fails on Windows if another process created `destination` first
+        temporary.renameTo(destination)
```
criticalAY: Can this fail?

criticalAY (commented):
I am ok with this change, just one question rest is good

rsdroid-testing/src/main/java/net/ankiweb/rsdroid/testing/FileUtils.kt
david-allison: Yes - security issues/disk space issues. Thanks!

A later `UnsatisfiedLinkError` would expose this, but I've fixed it to be more explicit

david-allison:
I plan to add 2/3 more commits here

Issue #674, opened 2026-04-07
Investigate converting `rsdroid[-testing]` to a `java-library`


david-allison:
* Speeds up tests
  * This is more useful now we're considering re-architecting the module structure of the app. If we can remove 'Robolectric' from a test suite for a module, we get huge savings. There are many tests which only use Robolectric due to the `rsdroid`/`libanki`, but we can't feasibly
* Fixes local builds: `Direct local .aar file dependencies are not supported when building an AAR.` 
    * https://github.com/ankidroid/Anki-Android-Backend/issues/671
    * This does not impact `androidTest`
* I was asked in the past about this for a university project. It seemed feasible at the time, but I didn't have capacity.

**Initial thoughts**

1. What are the blockers? (database; Looper etc...)
    * I do not believe these will cause significant problems. 
2. How will packaging the `.so` files work?
    * We currently have two artifacts: `rsdroid` with `.so` files for each of the 4 mainstream Android targets.
    * `rsdroid-testing` with  `rsdroid.dll`, `librsdroid.dylib` and `librsdroid.so` for Robolectric
        * Loaded via `RustBackendLoader`
3. Document `build/generated/anki_artifacts`
  * Used via `AssetManager` - mmap onto the apk

**Constraints**

* Bloating the apk size (fat JAR) is unacceptable
* Local `aar` -> `aar` references are unacceptable
  * `apk` -> local `aar` 
* `anki_artifacts` are heavily latency sensitive: Pages and web content used by the study screen

david-allison:
**How will packaging the .so files work?**

I've ideated with Claude, and I feel the best solution will be:

* `net.ankiweb:rsdroid:<v>` - `java-library` JAR - depended on by the dependency tree of `Anki-Android`
* `net.ankiweb:rsdroid-android:<v>` - AAR containing only the per-Android `.so` files and `anki_artifacts`. Depended on by the app module and `androidTest` tests
  * `anki_artifacts` requires an `AssetManager` to use the files, which is an Android dependency
    * TBC - is this going to cause a problem with aar dependencies
* `net.ankiweb:rsdroid-desktop:<v>` - `java-library` JAR containing `librsdroid.dylib` / `librsdroid.so` / `rsdroid.dll` + `BackendLoader`

david-allison:
Next: https://github.com/david-allison/Anki-Android-Backend/tree/java-library-8

PR #732, opened 2026-09-07
build(deps): migrate Timber -> SLF4J

david-allison:
* Part of #674

`slf4j-api` is pure JVM, so can be used in a `java-library`.

AnkiDroid uses com.arcao:slf4j-timber, so app log output is unchanged.

slf4j-api must stay on 1.7.x until AnkiDroid moves to a ServiceLoader-based provider (current unplanned).

Assisted-by: Claude Fable 5


criticalAY:
Looks good

PR #733, opened 2026-09-08
backend: reduce Android dependencies

david-allison:
> [!NOTE]
> Assisted-by: Claude Fable 5

Another set of extractions so we can move to a `java-library`

----

Recoding the main thread check using Reflection was possible, but it hasn't been triggered in years, and is effectively dead code now we have CollectionManager


BrayanDSO (approved):

criticalAY (approved):
Cool!

PR #741, opened 2026-09-16
build(deps): adopt anki 26.09.2

david-allison:
now `0.1.69-anki26.09.2`

https://github.com/ankitects/anki/releases/26.09.2

hash: bb0dd6d1310afe290dea5e5a440400f851789936

> [!NOTE]
> I have updated the below script to handle the case where the Windows installer template submodule was moved to a different GitHub repository.
> 
> My repo had the old URL cached, which caused an error on `git fetch`

```
cd anki
git fetch origin --recurse-submodules=no
git checkout 26.09.2
git submodule sync --recursive
git submodule update --init --recursive
cd ../
./build.sh
./check-rust.sh
```

* rust-toolchain.toml updated to 1.97.1
* `.truncate(0)` => `.clear()`


gradle.properties
```diff
@@ -19,7 +19,7 @@ android.useAndroidX=true
 android.enableJetifier=false
 
 GROUP=io.github.david-allison
-VERSION_NAME=0.1.68-anki26.05
+VERSION_NAME=0.1.69-anki26.09.2
```
mikehardy, line 22: oh! just loked and it appears we missed the .08 series and the first couple .09

mikehardy (approved):
LGTM

gradle.properties
david-allison: .03 will be out very soon, but I want to get the breaking changes in Anki-Android out the way

We should defer this for a very fast AnkiDroid 2.26 - 2.25 exists to keep us on the store

david-allison:
https://github.com/ankidroid/Anki-Android-Backend/actions/runs/35400065523

PR #742, opened 2026-09-16
Generate PreviewTranslations implementation

### LINKED ISSUE #740: Create static strings implementation of GeneratedTranslations to use in Android Studio composable preview mode
At this moment the backend strings can be accessed in composables in the Android app but they break the preview mode in Android Studio (which is important for quick iterations of composables UI) as the preview composables have certain restrictions which make the backend to not be initialized. As the general recommendation is to keep composable light we need another implementation for accessing backend strings which returns the strings directly.

This should be doable in _fluent.rs_ file(not going to pretend I know rust):

```rust
fn static_text_methods(modules: &[Module]) -> String {
    let mut out = vec![];
    for module in modules {
        for translation in &module.translations {
            let key = translation.key.replace('-', "_").to_camel_case();
            let arg_types = get_arg_types(&translation.variables);
            out.push(format!(
                "
    /** {} */
    fun {}({}): String {{
        return {}
    }}
",
                translation.text, key, arg_types, translation.text
            ));
        }
    }
    out.join("\n") + "\n"
}

fn_build_source() {
    ...
    out.push_str(&methods(modules));
    out.push('}');
    out.push_str("

interface StaticGeneratedTranslations {    
    ");
    out.push_str(&static_text_methods(modules))
    out.push('}');

    out
```

lukstbit:
I implemented the changes to generate the new PreviewTranslations implementation.

Closes  https://github.com/ankidroid/Anki-Android-Backend/issues/740

Note#1: Used ChatGPT for some of the rust code.
Note#2: ~Tried to use this(local library) with the original code and preview mode complains about missing GeneratedTranslations class...~ Works in the main module, not anki-common...


rslib-bridge/fluent.rs
```diff
@@ -50,6 +50,58 @@ fn methods(modules: &[Module]) -> String {
     out.join("\n") + "\n"
 }
 
+// A mapping of some of the parameters names passed to the generated strings functions and their proper valid kotlin
+// form. Needed because some of the parameters passed in do not match the Kotlin parameters names breaking the
+// generated code for PreviewTranslations.
+static KOTLIN_PARAMS_NAMES_MAPPING: &[(&str, &str)] = &[
+    ("$val", "$'val'"), // can't use val as a variable name in kotlin
```
david-allison: Can you use backticks?

rslib-bridge/fluent.rs
```diff
@@ -50,6 +50,58 @@ fn methods(modules: &[Module]) -> String {
...
+    ("$secs-per-card", "$secsPerCard"),
+];
+
+fn fix_incorrect_param_names(text: &str) -> String {
+    let mut result = text.to_string();
+    for (from, to) in KOTLIN_PARAMS_NAMES_MAPPING {
+        result = result.replace(from, to);
+    }
+    result
+}
```
david-allison: GPT-6 gave this, which also fixes an issue that Rust uses `{$var}` whereas kotlin uses `${var}`

```rs
  fn preview_text(translation: &Translation) -> String {
      translation.variables.iter().fold(
          translation.text.clone(),
          |text, arg| {
              text.replace(
                  &format!("{{${}}}", arg.name),
                  &format!("${{`{}`}}", arg.name.to_camel_case()),
              )
          },
      )
  }
```

EDIT: from Fable

```rs
/// Rewrites Fluent placeables (`{$deck_name}`) into Kotlin string templates
/// (`${`deckName`}`), using the same parameter names as `get_arg_types`.
fn kotlin_template_text(translation: &Translation) -> String {
    translation
        .variables
        .iter()
        .fold(translation.text.clone(), |text, var| {
            text.replace(
                &format!("{{${}}}", var.name),
                &format!("${{`{}`}}", var.name.to_camel_case()),
            )
        })
}
```

david-allison (changes requested):

rslib-bridge/fluent.rs
lukstbit: Much better, thanks.

I used the free version of chat gpt to navigate the rust syntax, but it's no match for all knowing Claude.

david-allison (approved):
LGTM

